PR #17, opened 2023-12-06
Add submodule promptsource

yhyu13:
Add prompt source as required by evaluation. Python version >=3.10 does not have packages available on pypi


tridao:
Can we just do `pip install git+https://github.com/bigscience-workshop/promptsource.git` instead of having promptsource as a submodule?

tridao:
I've just updated the README with instruction to install promptsource.

Issue #44, opened 2023-12-10
Error when `fused_add_norm` is False.


hyunwoongko:
```
llm-compute-14:   File "/home/azureuser/.conda/envs/megatron-deepspeed/lib/python3.8/site-packages/mamba_ssm/ops/triton/layernorm.py", line 494, in forward
llm-compute-14:     return rms_norm_fn(
llm-compute-14: TypeError: rms_norm_fn() got an unexpected keyword argument 'is_rms_norm'
```

The `rms_norm_fn` function doesn't have parameter named `is_rms_norm`.
Rather than, it's hard coded like the following.

https://github.com/state-spaces/mamba/blob/main/mamba_ssm/ops/triton/layernorm.py#L478

```python
def rms_norm_fn(x, weight, bias, residual=None, prenorm=False, residual_in_fp32=False, eps=1e-6):
    return LayerNormFn.apply(x, weight, bias, residual, eps, prenorm, residual_in_fp32, True)  <-- this
```

<br>

But, RMSNorm class in `ops/triton/layernorm.py` inputs `is_rms_norm` parameter to this function.

https://github.com/state-spaces/mamba/blob/main/mamba_ssm/ops/triton/layernorm.py#L502

```python
def forward(self, x, residual=None, prenorm=False, residual_in_fp32=False):
        return rms_norm_fn(
            x,
            self.weight,
            self.bias,
            residual=residual,
            eps=self.eps,
            prenorm=prenorm,
            residual_in_fp32=residual_in_fp32,
            is_rms_norm=True,  <-- this
        )
```

PR #45, opened 2023-12-10
Fix: remove is_rms_norm parameter in forward method of RMSNorm class

hyunwoongko:
This fixes https://github.com/state-spaces/mamba/issues/44.
Please refer to the issue to get more information about the bug.


tridao:
Thanks @hyunwoongko!

PR #47, opened 2023-12-11
Improve generation with some useful features

LuisLechugaRuiz:
This PR includes the following features:

- Add gitignore to avoid pushing compilation folders.
- Add save_pretrained (a minimal implementation) to avoid fetching from huggingface on start. This requires to store the configuration, I followed the same approach than Llama.
- Adds streamer to be able to stream tokens.
- Add explicit arguments to causal_conv1d_fn, [new version](https://github.com/Dao-AILab/causal-conv1d/commit/c06f47a372bc9c0806f1a17e58a8df153b9363c2) adds seq_idx as argument, we can solve compatibility with this simple trick.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+*__pycache__/
+*.egg-info/
+build/
+**.so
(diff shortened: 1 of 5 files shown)


tridao:
These are all great, thanks @LuisLechugaRuiz

PR #52, opened 2023-12-13
Implement bi-directionality

yair-schiff:
- [x] Implement bi-directionality by applying backbone twice: (1) to the forward sequence and (2) to the backward sequence.
- [x] Implement 3 strategies for combining forward / backward Mamba hidden states:
    1. `add`: Add the states.
    2. `concat`: Concatenate the states. This doubles the hidden dimension,`d_model`, which also prevents weight tying between `embedding` and `lm_head` weights.
    3. `ew_multiply`: perform element-wise multiplication between the states.

### DIFF
--- a/mamba_ssm/models/mixer_seq_simple.py
+++ b/mamba_ssm/models/mixer_seq_simple.py
@@ -2,6 +2,7 @@
 
 import math
 from functools import partial
+from typing import Union
 
 from collections import namedtuple
 
@@ -179,12 +180,20 @@ class MambaLMHeadModel(nn.Module, GenerationMixin):
         vocab_size: int,
         initializer_cfg=None,
         pad_vocab_size_multiple: int = 1,
+        bidirectional: bool = False,
+        bidirectional_strategy: Union[str, None] = None,
         device=None,
         dtype=None,
         **backbone_kwargs,
     ) -> None:
         factory_kwargs = {"device": device, "dtype": dtype}
         super().__init__()
+        if bidirectional and bidirectional_strategy not in ["add", "concat", "ew_multiply"]:
+            raise NotImplementedError(f"{bidirectional_strategy} strategy for bi-directionality is not implemented!")
+        if bidirectional and bidirectional_strategy is None:
+            bidirectional_strategy = "add"  # Default strategy: `add`
+        self.bidirectional = bidirectional
+        self.bidirectional_strategy = bidirectional_strategy
         if vocab_size % pad_vocab_size_multiple != 0:
             vocab_size += pad_vocab_size_multiple - (vocab_size % pad_vocab_size_multiple)
         self.backbone = MixerModel(
@@ -195,7 +204,10 @@ class MambaLMHeadModel(nn.Module, GenerationMixin):
             **backbone_kwargs,
             **factory_kwargs,
         )
-        self.lm_head = nn.Linear(d_model, vocab_size, bias=False, **factory_kwargs)
+        if bidirectional and bidirectional_strategy == "concat":
+            self.lm_head = nn.Linear(d_model * 2, vocab_size, bias=False, **factory_kwargs)
+        else:
+            self.lm_head = nn.Linear(d_model, vocab_size, bias=False, **factory_kwargs)
 
         # Initialize weights and apply final processing
         self.apply(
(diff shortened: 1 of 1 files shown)


mamba_ssm/models/mixer_seq_simple.py
```diff
@@ -83,6 +90,95 @@ def _init_weights(
...
+        hidden_states: (B, L, D)
+        Returns: same shape as hidden_states
+        """
+        out = self.mamba_fwd(hidden_states, inference_params=inference_params)
+        if self.bidirectional:
+            out_rev = self.mamba_rev(
+                hidden_states.flip(dims=(1,)),  # Flip along the sequence length dimension
+                inference_params=inference_params
+            ).flip(dims=(1,))  # Flip back for combining with forward hidden states
+            if self.bidirectional_strategy == "add":
```
Skylion007, line 175: bidirectional strategy should be an enum

mamba_ssm/models/mixer_seq_simple.py
```diff
@@ -234,8 +338,8 @@ def forward(self, input_ids, position_ids=None, inference_params=None, num_last_
         if num_last_tokens > 0:
             hidden_states = hidden_states[:, -num_last_tokens:]
         lm_logits = self.lm_head(hidden_states)
-        CausalLMOutput = namedtuple("CausalLMOutput", ["logits"])
-        return CausalLMOutput(logits=lm_logits)
+        LMOutput = namedtuple("LMOutput", ["logits"])
```
Skylion007, line 341: I know this isn't from your code, but constructing named tuples in this way is deprecated as it prevents proper type inference.

Skylion007 (commented):
Left some nits

sentialx:
What if the sequences have paddings? E.g.
Input is
[1 2 3 0 0 0]
So flipped input would be
[0 0 0 3 2 1].
Shouldn't it be
[3 2 1 0 0 0]?

yair-schiff:
@sentialx , agreed. That's a good catch.

jimmieliu:
how the speed compares to uni-directional?

yair-schiff:
> how the speed compares to uni-directional?

It's about 2x

pengzhangzhi:
@yair-schiff I am just curious, did you solve the 

> What if the sequences have paddings? E.g. Input is [1 2 3 0 0 0] So flipped input would be [0 0 0 3 2 1]. Shouldn't it be [3 2 1 0 0 0]?

Just curious, is this problem solved?

pengzhangzhi:
I came up with a solution to the padding issue. Say a tensor [1,2,3,0,0], where 0 is the padding token. We flip it to get [0,0,1,2,3], pass it to the network and flip it back. Therefore, the flipped tensor information matches the original tensor order as we apply double flips.
```
given: x
out = x + f(x.flip()).flip()
```

xuanwuji:
> I came up with a solution to the padding issue. Say a tensor [1,2,3,0,0], where 0 is the padding token. We flip it to get [0,0,1,2,3], pass it to the network and flip it back. Therefore, the flipped tensor information matches the original tensor order as we apply double flips.
> 
> ```
> given: x
> out = x + f(x.flip()).flip()
> ```

Hi,  Your approach is clever! But I have a question: if you flip the input to [0,0,1,2,3], does the padding in front of it affect sequence hidden features learning? i.e., does it produce a different result(bad repersentation of sequence) than the input of [3,2,1,0,0]?
I don't know enough about it, could you possibly give me some guidance? This will help me a lot. Thank you very much!

Museum7432:
@xuanwuji well, you can remove the leading paddings by shifting each row of x before flipping x. As for its effect, since the hidden state is initialized with 0, it should still be filled with 0 after scanning through the paddings. So, those padding shouldn't have any effect on the result. However, you can use the following function just to be sure.
```
def flip_padded_hidden_states(hidden_states, seq_lens):
    batch_size, seq_len, hidden_dim = hidden_states.shape

    indices = torch.arange(batch_size * seq_len, device=hidden_states.device).reshape(
        batch_size, seq_len
    )

    indices_offset = seq_len - seq_lens

    indices = (indices - indices_offset.unsqueeze(1)) % (seq_len * batch_size)

    indices = indices.flip(1)

    return hidden_states.reshape(batch_size * seq_len, hidden_dim)[indices]
```
checking the effect of paddings:
```
import torch
from mamba_ssm import Mamba2, Mamba
from torch.nn import functional as F

batch, length, dim = 2, 64, 16

model = Mamba(
    d_model=dim, # Model dimension d_model
    d_state=16,  # SSM state expansion factor
    d_conv=4,    # Local convolution width
    expand=2,    # Block expansion factor
).to("cuda")

x = torch.randn(batch, length, dim).to("cuda")
padded_x = F.pad(x, (0,0, 4,0))

y = model(x)
padded_y = model(padded_x)

unpadded_y = padded_y[:,4:]

print(f'Output max diff: {(unpadded_y - y).abs().max().item()}')
print(f'Output mean diff: {(unpadded_y - y).abs().mean().item()}')
```

PR #83, opened 2023-12-27
Update __init__.py

Koziev:
Make MambaConfig class available from outside in order to allow MambaLMHeadModel customization via constructor ```config``` argument.


Koziev:
1) Minor changes to make MambaLMHeadModel more transformers.Trainer-friendly: return loss as a first element from ```forward```.
2) Adding filter to config keys in ```from_pretrained``` so config.json can have fields ```_name_or_path``` and ```architectures``` like other models on hf.

RonanKMcGovern:
Would be great if this could be merged and included in a pre-built wheel.

PR #103, opened 2024-01-15
BUGFIX: layernorm mean, rstd

gaxler:
Layernorm kernel fails if your tensors are not on default cuda device.

mean and rstd were allocated on default "cuda" device.


tridao:
Thanks @gaxler!

Issue #27, opened 2023-12-07
how to compare mamba with flashattention2


xiayuqing0622:
In your paper, you mentioned that mamba scan is faster than flashattention2.  
Does it mean comparing https://github.com/state-spaces/mamba/blob/0131c1e94a46fc9f70bcfc9d57962963bb2f0b9e/mamba_ssm/ops/selective_scan_interface.py#L14 with https://github.com/Dao-AILab/flash-attention/blob/9356a1c0389660d7e231ff3163c1ac17d9e3824a/flash_attn/flash_attn_interface.py#L432 ? 
The inputs of these two modules are different, is this comparation fair? Or the preprocessing(compute q, k, v in flashattention; compute A,B,C,D,delta in mamba scan) need to be be taken into account?

tridao:
We compared attention time (softmax(QK^T)V) vs scan time, without the linear projection.
The dimensions are different, e.g. in a 1.3B model Transformers would typically have Q, K, V of hidden dimensions 2048, while the input to the scan would be of dimension 4096 (due to expand=2). So we measured attention with QKV of dim 2048 and scan with input of dimension 4096 to be fair / favorable to attention.

xiayuqing0622:
> We compared attention time (softmax(QK^T)V) vs scan time, without the linear projection. The dimensions are different, e.g. in a 1.3B model Transformers would typically have Q, K, V of hidden dimensions 2048, while the input to the scan would be of dimension 4096 (due to expand=2). So we measured attention with Q, K, V of dim 2048 and scan with input of dimension 4096 to be fair / favorable to attention.

And what datatype did you use?

tridao:
Q, K, V are bf16 for attention.
u, delta, B, C, z are bf16, A and D are fp32 for scan.

xiayuqing0622:
> Q, K, V are bf16 for attention. u, delta, B, C, z are bf16, A and D are fp32 for scan.

it works now, thank you!

xiayuqing0622:
> Q, K, V are bf16 for attention. u, delta, B, C, z are bf16, A and D are fp32 for scan.

[...]

PR #161, opened 2024-02-04
add feature / fix bug: I fixed the `kNRows` feature in forward

MzeroMiko:
Thank you for sharing this splendid work!

I found that `kNRows` is always `1` in `original selective_scan`, and I observed that if I use greater `kNRows` in `selective scan`, the faster the code would run. The phenomenon is consistent with [mamba.py](https://github.com/alxndrTL/mamba.py/tree/main), when adding `d_state`, the time consumption keeps. Though it is not strictly right, but adding the burden of one thread and reducing the number of blocks (as SM is limited) really works in most of [cases](https://github.com/alxndrTL/mamba.py/issues/8). 

So I reopen that feature which may be deprecated in `original selective_scan`, and fixed some bugs related to it.
I have tested with `pytest tests/ops/test_selective_scan_.py` (which you may delete later), and all tests pass.

Note that I have only fixed the forward procedure, so in backward, `nrows` is still `1`.

Before Merging: I found that, when I uncomment all alternative parameters, the test is not all pass. However, `mamba_ssm-1.1.3.post1+cu122torch2.2cxx11abiFALSE-cp310-cp310-linux_x86_64.whl` acts the same.


apoorv2904:
@MzeroMiko could you please share you benchmark numbers and platform. These are the times I get on A100 80GB GPU
image

My benchmark code can be found here
https://github.com/state-spaces/mamba/issues/27#issuecomment-1930747882

MzeroMiko:
Thank you very much, @apoorv2904. 
You are right, and I nearly failed to reproduce the results I have observed before.
These days, I have been working on it. (the environment I use is 4090 24G, with py310+cu121+torch2.2)
1. I added `nrow` feature in backward to better compare with different `nrow` settings.
2. I compared my code (`selective_scan_test` here, or `selective_scan_core` in VMamba) with `mamba_ssm` rather than `selective_scan_ref`, and keeps no difference (tested all pass with [test file](https://github.com/MzeroMiko/mamba/blob/main/kernel/test_selective_scan_new2old.py)).
3. I realised that the [issue]`https://github.com/alxndrTL/mamba.py/issues/8` proves nothing here, since raising `d_state` only inference the flops in SSM (nearly equals selective scan) while raising `d_model` or `seqlen` inferences the whole mamba model. As SSM is fast compared to `the whole model + data loading`, the speed difference is small and hard to observe (which is one possibility to that issue).
4. I used my newly written [`simple benchmark`](https://github.com/MzeroMiko/mamba/blob/main/kernel/test_selective_scan_speed.py), and found the results are consistent with yours. It seems that raissing nrows would only make the code slower, until I finally realised that ***the test which shows raising the nrow will raise the speed, was done in 7x7 images, which means seqlen is 49! extremely small! ***. Then I add `seqlen=64` in testing, and found in some `fwdnrow+bwdnrow` patterns, the speed is fast, see [log](https://github.com/MzeroMiko/mamba/blob/main/kernel/test_selective_scan_speed.log) for details. Though I still do not know how bwd codes inferences the fwd procedure.
5. I modified your [`benchmark`](https://github.com/MzeroMiko/mamba/blob/main/kernel/test_selective_scan_benchmark.py), and the results are consistent with `test_selective_scan_speed`, see [log](https://github.com/MzeroMiko/mamba/blob/main/kernel/test_selective_scan_benchmark.log) for details.
To conclude, with short `seqlen`, bigger `nrow` may lead to faster speed, but the reason remains unknown.

PR #168, opened 2024-02-07
Bugfix `causal_conv1d_fn` interface

HarborYuan:
Hi @tridao ,

Thanks for your great work on mamba.

I noticed a potential bug in the version upgrade regarding the causal_conv1d_fn interface. Could you please help me check whether the fix is correct?

Thanks for your attention.

Best Regards,
Haobo


tridao:
Thanks!

PR #185, opened 2024-02-23
Fix directory creating for saving with multiprocessing

deroholic:
The save_pretrained() method can and often will fail when training with multi-processes because the save directory can be created by another process after the check of its existence which will result in an exception from os.makedirs().


tridao:
Thanks!

PR #188, opened 2024-02-23
Allow the model to be trained with most frameworks.

deroholic:
If labels are passed to the model along with the input_ids, it will return the loss.  Otherwise it functions the same as it did originally.  This allows it to be used for training as-is with most frameworks.


mamba_ssm/models/mixer_seq_simple.py
```diff
@@ -225,7 +225,7 @@ def tie_weights(self):
     def allocate_inference_cache(self, batch_size, max_seqlen, dtype=None, **kwargs):
         return self.backbone.allocate_inference_cache(batch_size, max_seqlen, dtype=dtype, **kwargs)
 
-    def forward(self, input_ids, position_ids=None, inference_params=None, num_last_tokens=0):
+    def forward(self, input_ids, position_ids=None, inference_params=None, num_last_tokens=0, attention_mask=None, labels=None):
```
zigzagcai, line 228: I am curious about this implementation but I noticed that `attention_mask` is not used, so how to avoid cross contamination?

PR #211, opened 2024-03-01
Don't hardcode tied embed+unembed

haileyschoelkopf:
I encountered problems when trying to use a Mamba model trained with untied embed + unembed weights in this codebase because `MambaLMHeadModel.tie_weights()` is always called.

This PR adds a `tie_embeddings` field to the config which defaults to true, and only ties the input + output embeddings if the config value is true.


tridao:
Thanks!

PR #265, opened 2024-03-24
selective_scan_cuda error

wang935415150:
I'm using the m1 chip version of MacOS and python3.10 pytorch2.2.1 natively tried to use mamba_ssm.ops.selective_scan_interface native, so I tried to skip here, the truth is that it works, and it can also call model.to ("mps") so I made this modification attempt

### DIFF
--- a/mamba_ssm/ops/selective_scan_interface.py
+++ b/mamba_ssm/ops/selective_scan_interface.py
@@ -13,7 +13,10 @@ except ImportError:
     causal_conv1d_fn = None
     causal_conv1d_cuda = None
 
-import selective_scan_cuda
+try:
+    import selective_scan_cuda
+except ImportError:
+    selective_scan_cuda = None
 
 
 class SelectiveScanFn(torch.autograd.Function):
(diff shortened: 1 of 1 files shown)


radarFudan:
This is an interesting discovery. Just curious: is there a significant speedup from mps over cpu?

tridao:
You can put the import in the try except, but I wouldn't call the `selective_scan_ref` function in `selective_scan_fn` if `selective_scan_cuda` is not found. Instead it should error.
We don't want people to silently get much slower performance if they forgot to install the CUDA extension, or the installation was not correct.

wang935415150:
> 这是一个有趣的发现。只是好奇：mps 相对于 cpu 是否有显着的加速？
Hello, my English is not very good, so I took the translation tool and replied: You can check out this official document, in fact he has some improvements.
https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

wang935415150:
> This is an interesting discovery. Just curious: is there a significant speedup from mps over cpu?

Hello, my English is not very good, so I took the translation tool and replied: You can check out this official document, in fact he has some improvements.
https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/

DowneyFlyfan:
> I'm using the m1 chip version of MacOS and python3.10 pytorch2.2.1 natively tried to use mamba_ssm.ops.selective_scan_interface native, so I tried to skip here, the truth is that it works, and it can also call model.to ("mps") so I made this modification attempt

Can you give me a brief intro about how it works on mps device ? I would appreciate it if you can contact me

PR #335, opened 2024-05-27
Square plus instead of Softplus

kashif:
Add a flag to the mamba kernels to use squareplus instead of softplus to avoid log/exps

- [ ] tests failing


albertfgu:
Have you found this change to help anywhere?

kashif:
thanks @albertfgu I am gonna test it out next... happy to move this PR To draft?

kashif:
my hypothesis is that it should be faster... but i do not expect this PR to learn better representations... I am currently trying to get the tests passing though

PR #359, opened 2024-06-05
Adding ROCm support [AMD official]

amoskvic:
The pull request includes 

- Adjustments to CUDA kernel code to make it hipify-friendly (i.e. make it possible for PyTorch hipification routine to automatically create .hip source code when the package is built on ROCm). The original CUDA code was kept intact, except for minor changes that do not affect behavior (e.g. adding "typename" for cases when hip compiler fails to infer them). All ROCm-specific adjustments are wrapped in #ifdef/#ifndef USE_ROCM.
- Minor adjustments to python code. Mostly in setup.py. Again, we aimed to make the code behave exactly the same on CUDA machines. We also needed to make small changes around the triton autotune routine to account for different hardware specs (warp size).
- Minor adjustments to the readme file (ROCm 6.0 requires a quick patch for the build to work).

The push does not include:
- ROCm-specific optimizations. We identified a few cases where performance on ROCm machines can be improved, but this pull request is primarily for functional support.
- Machinery to release ROCm-specific package builds (we haven't touched the github workflow file as of yet).


stephen-youn:
there is another PR that makes mamba's triton kernels be compatible with the latest triton 3.0.0 so that both mamba and mamba2 can run  with the ROCm (#377)

PR #385, opened 2024-06-11
Fix typo in min_p sampling

epicfilemcnulty:
Hey guys, quite a while ago I made a PR for min_p sampling support in Mamba,
and it was merged. 

Unfortunately, there was a small mistake in the code, which
I've discovered only later and forgot to create a PR with the fix. Sorry about that, my bad. 

News of Mamba2 release reminded me about the issue, so here is the fix.


tridao:
Thanks @epicfilemcnulty

PR #402, opened 2024-06-18
Make properly pip installable without needing build isolation (uv pip install now works too)

kszlim:
Now you can just pip install mamba_ssm and it will work, previously it would throw unless you had `packaging` and `torch` already installed. It would also always fail if you had build isolation on (which will be on by default in the future within `pip`).

As a side effect, now `uv` can `uv pip install mamba_ssm` properly now.

Take note that installing `causal-conv1d` via the optional extra will still fail with build isolation due to that package needing similar changes to this one.


kdre:
Please approve this pr. It also fixed my installation issue.

amrit110:
Hi @tridao when is a new release to pypi expected with this fix? The installation of the latest available package fails.

PR #462, opened 2024-07-13
repeat kv for GQA

lchu-ibm:
It seems currently we are not repeating kv when GQA is used. 

This PR fix this issue by adding kv repeat before passing kv to SDPA.


lchu-ibm:
@tridao

PR #504, opened 2024-07-30
add support for pytorch 2.4.0 with cuda 12.4

winglian:
Update wheels publishing now that pytorch 2.4.0 is out and it also seems to support CUDA 12.4 too
Screenshot 2024-07-30 at 10 50 45 AM


winglian:
@tridao Could you also re-trigger the publish wheels action for this so the 2.4.0 wheels get built please?

Issue #503, opened 2024-07-30
CUDA error when using Mamba2 with long context


titzehong:
Hi, I am benchmarking inference speed on long sequences and encountering CUDA-related errors specifically with the Mamba2 models at longer sequence lengths (>200k). This issue does not occur with Mamba1 models.

For example running:
```
python benchmark_generation_mamba_simple.py --model-name "state-spaces/mamba2-1.3b" --promptlen 300000 --genlen 1
```
produces the error:

  File "/usr/local/lib/python3.10/dist-packages/mamba_ssm/utils/generation.py", line 260, in generate
    output = decode(
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/mamba_ssm/utils/generation.py", line 221, in decode
    scores.append(get_logits(sequences[-1], inference_params))
  File "/usr/local/lib/python3.10/dist-packages/mamba_ssm/utils/generation.py", line 184, in get_logits
    logits = model(
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/mamba_ssm/models/mixer_seq_simple.py", line 279, in forward
    hidden_states = self.backbone(input_ids, inference_params=inference_params, **mixer_kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/mamba_ssm/models/mixer_seq_simple.py", line 194, in forward
    hidden_states, residual = layer(
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1520, in _call_impl
[...]

PR #537, opened 2024-08-24
Fix Incorrect Gradients and Illegal Memory Access Error in Mamba2

Hprairie:
Hey Tri and Albert,

Digging through the code base, I have found a bug in the backward pass for the gradient calculations in the function `_chunk_scan_bwd_ddAcs_stable_kernel`. I think this solves the issue given by #503, however, it will definitely fix incorrect gradient calculations for $\Delta A$.

Essentially, all that was going on lines [1146-1149](https://github.com/Hprairie/mamba/blob/b069d47659435f613d3820d6f5a5e188ad46a4f2/mamba_ssm/ops/triton/ssd_chunk_scan.py#L1146) if `BLOCK_SIZE_N` was greater then `BLOCK_SIZE_M` then it would update the pointer by `BLOCK_SIZE_N`. Then when we try to zero out the remainder of the sequence in lines [1152:1154](https://github.com/Hprairie/mamba/blob/b069d47659435f613d3820d6f5a5e188ad46a4f2/mamba_ssm/ops/triton/ssd_chunk_scan.py#L1152), the loop assumes that pointer of `ddA_cs_ptr` is at `hi=(pid_m + 1) * BLOCK_SIZE_M` when we could have overshot it, given that `BLOCK_SIZE_M < BLOCK_SIZE_N`. 

All that I do in the PR is remove the potential for us to overshoot by disabling the kernel configs where `BLOCK_SIZE_N > BLOCK_SIZE_M`. This means that at the end of the first loop we will always have out ptr updated by `(pid_m + ) * BLOCK_SIZE_M`. More complicated things can be done to fix this, however, I don't want to add complexity to the kernel, so I think that this works for now.

I found that incorrect gradient calculations for `dA` were incredibly common when this bug was present, thus I don't know how much this will have affected model training.

Let me know if you need any clarifications or if I can do anything else to help!

### DIFF
--- a/mamba_ssm/ops/triton/ssd_chunk_scan.py
+++ b/mamba_ssm/ops/triton/ssd_chunk_scan.py
@@ -1133,7 +1133,7 @@ def _chunk_scan_bwd_ddAcs_stable_kernel(
         # If there's seq_idx, we already zero'ed out cb[i, j] for seq_idx[i] != seq_idx[j]
         cb = tl.load(cb_ptrs, mask=(offs_m[:, None] < chunk_size) & (offs_n[None, :] < chunk_size - start_n), other=0.0).to(tl.float32)
         acc *= cb
-        dA_cs_n = tl.load(dA_cumsum_ptr + start_n + offs_n * stride_dA_cs_csize, mask=offs_n < chunk_size - start_n, other=0.0).to(tl.float32)
+        dA_cs_n = tl.load(dA_cumsum_ptr + (start_n + offs_n) * stride_dA_cs_csize, mask=offs_n < chunk_size - start_n, other=0.0).to(tl.float32)
         acc *= tl.exp(dA_cs_m[:, None] - dA_cs_n[None, :])
         mask = offs_m[:, None] >= start_n + offs_n[None, :] + 1
         acc = tl.where(mask, acc, 0.0)
--- a/mamba_ssm/ops/triton/ssd_chunk_scan.py
+++ b/mamba_ssm/ops/triton/ssd_chunk_scan.py
@@ -1055,11 +1055,11 @@ def _chunk_scan_bwd_ddAcs_stable_kernel_old(
 @triton.autotune(
     configs=[
         triton.Config({'BLOCK_SIZE_M': 32, 'BLOCK_SIZE_N': 32}, num_stages=3, num_warps=4),
-        triton.Config({'BLOCK_SIZE_M': 32, 'BLOCK_SIZE_N': 64}, num_stages=3, num_warps=4),
-        triton.Config({'BLOCK_SIZE_M': 32, 'BLOCK_SIZE_N': 128}, num_stages=3, num_warps=4),
+        # triton.Config({'BLOCK_SIZE_M': 32, 'BLOCK_SIZE_N': 64}, num_stages=3, num_warps=4),
+        # triton.Config({'BLOCK_SIZE_M': 32, 'BLOCK_SIZE_N': 128}, num_stages=3, num_warps=4),
         triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 32}, num_stages=3, num_warps=4),
         triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 64}, num_stages=3, num_warps=4),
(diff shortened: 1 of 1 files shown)


Hprairie:
@tridao @albertfgu , I'm also going to ping you guys due to the nature of the problem

Hprairie:
Also, I should mention that for the incorrect gradients, it will happen when we accidently zero out already computed and stored gradients by other thread blocks.

tridao:
Thanks so much @Hprairie! Let me take a careful look

Hprairie:
Also fixed another bug in `_chunk_scan_bwd_ddAcs_stable_kernel`, in line [1136](https://github.com/state-spaces/mamba/blob/62db608da60f6fc790b8ed9f4b3225e95ca15fde/mamba_ssm/ops/triton/ssd_chunk_scan.py#L1136) we also don't multiply `start_n` by the stride.

tridao:
@Hprairie is it simpler to set `ddA_cs_ptrs` to `hi` after the loop? Then we don't need to worry about which block M & block N would work?

Hprairie:
@tridao Yes, that would be simpler I think, the only problem is that you need to revert the pointer updates being done in the first loop which either requires saving the original pointers or reverting the update, which is not as simple. Another thing that could be done (which should fix the problem) is to do `new_hi = tl.cdiv(hi, BLOCK_SIZE_N) * BLOCK_SIZE_N)` after the first loop. You could also only do this computation if `BLOCK_SIZE_N > BLOCK_SIZE_M` which the compiler should be able to figure out.  It's been a while since I tried this but it might have been slower bc of poor triton optimizations. I don't have that much bandwidth rn to try, but was going to toy around with some new idea for mamba kernels in a month and could try some stuff then

tridao:
Oh i was gonna do the simple thing.
Instead of 
```
    for start_n in range(hi, chunk_size, BLOCK_SIZE_N):
        tl.store(ddAcs_ptrs + stride_ddA_cs_csize_n, tl.zeros((BLOCK_SIZE_N,), dtype=tl.float32), mask=offs_n < chunk_size - start_n - 1)
        ddAcs_ptrs += BLOCK_SIZE_N * stride_ddA_cs_csize_n
```
We can do
```
    for start_n in range(hi, chunk_size, BLOCK_SIZE_N):
        tl.store(ddA_cumsum_ptr + (offs_n + start_n + 1) * stride_ddA_cs_csize_n, tl.zeros((BLOCK_SIZE_N,), dtype=tl.float32), mask=offs_n < chunk_size - start_n - 1)
```

Hprairie:
Ahh lol yeah that seems much simpler. Also, I could be wrong, but I don't think the `+1` is necessary.

klae01:
I trained models using this repository's implementation **prior to this PR** and consistently observed a ~0.2% difference in grad norm across runs starting from the same checkpoint and data. This raises concerns about non-reproducible training behavior.

Does this fix significantly impact gradient backpropagation? If so, it might suggest that Mamba2 models trained with the pre-PR implementation were affected. Addressing this soon would be valuable, as future SSM research may build upon this repository.


grad norm:
image
loss:
image

> Oh i was gonna do the simple thing. Instead of
> 
> ```
>     for start_n in range(hi, chunk_size, BLOCK_SIZE_N):
>         tl.store(ddAcs_ptrs + stride_ddA_cs_csize_n, tl.zeros((BLOCK_SIZE_N,), dtype=tl.float32), mask=offs_n < chunk_size - start_n - 1)
>         ddAcs_ptrs += BLOCK_SIZE_N * stride_ddA_cs_csize_n
> ```
> 
> We can do
> 
> ```
>     for start_n in range(hi, chunk_size, BLOCK_SIZE_N):
>         tl.store(ddA_cumsum_ptr + (offs_n + start_n + 1) * stride_ddA_cs_csize_n, tl.zeros((BLOCK_SIZE_N,), dtype=tl.float32), mask=offs_n < chunk_size - start_n - 1)
> ```

Hprairie:
It's hard to say for sure. The current branch of the repo should be fine as the pr was merged. Before the merge, there was a chance that models trained had wrong gradient calculations, but it is hard to say exactly because it depends on which `config` triton's autotuner chooses. If it chooses a config that I have removed in the PR, then incorrect gradients will likely occur due to non-deterministic GPU scheduling. 

The hardest part about the bug was that it was probabilistic, meaning that running the code multiple times may/may not have it occur. 

Thus it's hard to say if this fix will have an impact or not on your specific machine, but there is a non-zero probability that it will.

PR #560, opened 2024-09-10
[Deprecation] update deprecated pytorch `custom_fwd` and `custom_bwd` function from `torch.cuda.amp`

Jonathanjordan21:


Jonathanjordan21:
@vasqu oh right, I didnt see it. Thanks

PR #596, opened 2024-10-18
Fix custom fwd and bwd for older PyTorch versions

vasqu:
Fixes #594
Fixes #496 
Fixes #517

#584 broke main for torch with older versions, e.g. torch==2.2.0. There have been similar attempts at #501 and #560
The main difference is to create a wrapper around the decorator that passes the new kwarg if necessary and changes the import depending on what torch version we have (based on if amp has custom_(fwd|bwd)). I moved it to the utils folder, open to change the structure.

P.S. Verified the fix with torch=2.2.0 (older version who doesn't have custom_(fwd|bwd) in amp) and torch=2.5.0 (newer version who does have custom_(fwd|bwd) in amp).

PR #608, opened 2024-10-26
Fix custom fwd and bwd for older PyTorch versions

KokeCacao:
Related to #596 
The fix works okay and tested on my own project.
See https://github.com/state-spaces/mamba/pull/596#issuecomment-2439662703

### DIFF
--- a/mamba_ssm/utils/torch.py
+++ b/mamba_ssm/utils/torch.py
@@ -1,16 +1,18 @@
 import torch
 from functools import partial
+from typing import Callable
 
-
-def custom_amp_decorator(dec, cuda_amp_deprecated):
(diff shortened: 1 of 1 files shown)


saurabh-kataria:
Hi, I had just opened an issue about this (https://github.com/state-spaces/mamba/issues/609). Your solution works. Thanks.

vasqu:
Sry about that! And thx for fixing it properly this time

mamba_ssm/utils/torch.py
```diff
@@ -1,16 +1,18 @@
 import torch
 from functools import partial
```
vasqu, line 2: We can safely remove this now :)

Issue #571, opened 2024-09-21
Significant differences in gradients between `_ref` and `_fn` when using the complex formulation.


karannb:
Hi, I was using complex dynamics for an application, and was seeing large differences in gradients computed by `mamba_inner_ref` and `mamba_inner_fn`, the scan functions worked fine and performed much better in my test (however, even for that case I had to lower my tolerance to 1e-6 from 1e-8 for the real case). I am attaching a reproducible sample below for the test for `mamba_inner_ref`, I am assuming this happens because of torch-with-complex-numbers is still under development, but would appreciate any guidance on how to solve this.

```python
'''
Check for mamba_inner_fn
'''
import math
import torch
from torch import nn
from tqdm import tqdm
from einops import repeat
import torch.nn.functional as F
from mamba_ssm.ops.selective_scan_interface import mamba_inner_fn, mamba_inner_ref

# Define a random seed for reproducibility
torch.manual_seed(0)
torch.cuda.manual_seed(0)

# Set device to CUDA if available, otherwise CPU
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

def test_gradient_implementation(device=device):
    # Create random input tensors and parameters
    batch_size = 4
    dstate = 10
    dim = 3
    seqlen = 7

    xz = torch.randn(batch_size, dstate*2, seqlen, device=device, requires_grad=True)
    conv1d_weight = torch.randn(dstate, 1, 4, device=device, requires_grad=True)
    conv1d_bias = torch.randn(dstate, device=device, requires_grad=True)
    x_proj_weight = torch.randn(dim*4 + 1, dstate, device=device, requires_grad=True)
    dt_proj_weight = torch.randn(dstate, 1, device=device, requires_grad=True)
    # Initialize special dt projection to preserve variance at initialization
    dt_init_std = 1**-0.5
    nn.init.uniform_(dt_proj_weight, -dt_init_std, dt_init_std)
    dt_bias = torch.randn(dstate, device=device, requires_grad=True)
    # Initialize dt bias so that F.softplus(dt_bias) is between dt_min and dt_max
    dt = torch.exp(
        torch.rand(dstate) * (math.log(0.1) - math.log(0.001))
        + math.log(0.001)
    ).clamp(min=1e-4)
    # Inverse of softplus: https://github.com/pytorch/pytorch/issues/72759
    inv_dt = dt + torch.log(-torch.expm1(-dt))
    with torch.no_grad():
[...]

Issue #137, opened 2024-01-29
Question Regarding Randomness


wyhsleep:
Hello,

Thanks for your interesting work but I have a question about the code that I'd like to discuss with you. Despite fixing all the random seeds, I'm still observing randomness in the results of my runs.
And I use the`torch.use_deterministic_algorithms(True)` and run the example code as follows:
`mport torch
from mamba_ssm import Mamba
torch.use_deterministic_algorithms(True)

batch, length, dim = 2, 64, 16
x = torch.randn(batch, length, dim).to("cuda")
model = Mamba(
    # This module uses roughly 3 * expand * d_model^2 parameters
    d_model=dim, # Model dimension d_model
    d_state=16,  # SSM state expansion factor
    d_conv=4,    # Local convolution width
    expand=2,    # Block expansion factor
).to("cuda")
y = model(x)
assert y.shape == x.shape`

I got the error message:
`Traceback (most recent call last):
  File "test.py", line 14, in <module>
    y = model(x)
  File "/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1501, in _call_impl
    return forward_call(*args, **kwargs)
  File "//lib/python3.10/site-packages/mamba_ssm/modules/mamba_simple.py", line 136, in forward
    self.in_proj.weight @ rearrange(hidden_states, "b l d -> d (b l)"),
RuntimeError: Deterministic behavior was enabled with either `torch.use_deterministic_algorithms(True)` or `at::Context::setDeterministicAlgorithms(true)`, but this operation is not deterministic because it uses CuBLAS and you have CUDA >= 10.2. To enable deterministic behavior in this case, you must set an environment variable before running your PyTorch application: CUBLAS_WORKSPACE_CONFIG=:4096:8 or CUBLAS_WORKSPACE_CONFIG=:16:8. For more information, go to https://docs.nvidia.com/cuda/cublas/index.html#cublasApi_reproducibility`
It seems that there is a randomness in the mamba module. Did you guys encounter this before?
Thank you for your help and thank you for your great job again!

tridao:
Did you follow the suggestion in the error message?

wyhsleep:
Yes, but the randomness still remains

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

wyhsleep:
[...]

PR #670, opened 2025-01-14
add `mamba_chunk_scan_combined` and `mamba_split_conv1d_scan_combined` tests

garrett361:
This PR adds correctness tests for `mamba_chunk_scan_combined` and `mamba_split_conv1d_scan_combined`, which seemed to be missing.  Forwards and backwards are tested against their reference implementations. Correctness when providing `seq_idx` is also tested.


garrett361:
@tridao I know the kernels inside of `mamba_chunk_scan_combined` and `mamba_split_conv1d_scan_combined` are individually tested, but I thought it would be worth it to add these more end-to-end tests. Thoughts?p

peterbjorgensen:
Any idea why the tolerances need to be that high?
Those tolerances seem very high for float32.
It is probably related to #683  #571

garrett361:
Yes, concerningly high, at least for the backwards where some tests need `tol = 1e-1` and/or are sensitive to seeds.

My first suspicion was that it is an issue with the tests, rather than the kernels, but I haven't found any problems yet.  And since the forwards tests pass at reasonable-ish 1e-2/1e-3 levels, any error would need to be a bit subtle. 

I have also found some non-determinism with the backwards passes for the `D` grads. Haven't posted about it yet; will try to today.

garrett361:
Also, also this is relevant: non-determinism is expected in the backwards due to [atomic adds](https://github.com/state-spaces/mamba/issues/137#issuecomment-1918483734), apparently.

karannb:
> Any idea why the tolerances need to be that high? Those tolerances seem very high for float32. It is probably related to #683 #571

Hi, thanks for mentioning this. I posted a solution for my case in #571 , you might want to check that. I was able to manage tolerances upto 1e-8 for all gradients and outputs.

PR #713, opened 2025-03-27
Add numerical instability patch and enable nheads to be a non multiple of 8 

rogerwaleffe:
This PR adds the following:

1. Differences of negative numbers that should be negative are clamped to zero. In most cases this clamping has no effect. In the case where numerical error leads the difference to be positive, this clamping prevents a positive number from being exponentiated leading to infinities/nans.
2. xBC is made contiguous before causal_conv1d to prevent a stride error (`RuntimeError: causal_conv1d with channel last layout requires strides (x.stride(0) and x.stride(2)) to be multiples of 8`) from being thrown when the number of heads is not a multiple of 8. With this patch, the number of heads can be any positive integer.
3. A ddt -> dt typo is fixed


tridao:
I think `contiguous` should only be called if the stride is not a multiple of 8. In other cases calling `contiguous` would incur an extra kernel?

rogerwaleffe:
Yeah I agree. I'll add that

peterbjorgensen:
This patch only fixes the training path (triton ops), but the numerical issues remain in the inference paths.

PR #727, opened 2025-05-17
update published wheels up to torch 2.7.0 and cuda 12.8.1

winglian:


sysradium:
Is something stopping us from this getting merged?

benglewis:
It looks like #810 replaces this

tridao:
The newest wheels support this now

PR #778, opened 2025-08-27
Update setup.py

naruto110:
To adapt to the RTX 5090, we have modified the **setup.py** file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -192,8 +192,6 @@ if not SKIP_CUDA_BUILD:
             cc_flag.append("arch=compute_90,code=sm_90")
         if bare_metal_version >= Version("12.8"):
             cc_flag.append("-gencode")
-            cc_flag.append("arch=compute_100,code=sm_100")
-            cc_flag.append("-gencode")
             cc_flag.append("arch=compute_120,code=sm_120")
 
 


tridao:
rtx 5090 is sm120 and it's already included. Why would removing sm100 help?

johnnynunez:
> rtx 5090 is sm120 and it's already included. Why would removing sm100 help?

could you add thor and spark?
11.0 and 12.1?

also mamba must be adapted to cuda 13

naruto110:
> rtx 5090 is sm120 and it's already included. Why would removing sm100 help?

Yes, the setup file has indeed already included it. However, after compiling and installing, although the package can be imported, the RTX 5090 cannot be properly initialized to train a model. I suspect this is because the default library compiled for **sm100** is being used. Once I removed **sm100**, the 5090 was able to start training normally.

johnnynunez:
> > rtx 5090 is sm120 and it's already included. Why would removing sm100 help?
> 
> Yes, the setup file has indeed already included it. However, after compiling and installing, although the package can be imported, the RTX 5090 cannot be properly initialized to train a model. I suspect this is because the default library compiled for **sm100** is being used. Once I removed **sm100**, the 5090 was able to start training normally.

i don't have problems with actual setup:
```bash
git clone --depth=1 --recursive https://github.com/state-spaces/mamba  /opt/mamba

# Navigate to the directory containing mamba's setup.py
cd /opt/mamba

MAX_JOBS="$(nproc)" \
MAMBA_FORCE_BUILD="TRUE" \
MAMBA_SKIP_CUDA_BUILD="FALSE" \
python3 setup.py bdist_wheel --dist-dir=/opt/mamba/wheels
pip3 install /opt/mamba/wheels/mamba*.whl
```

naruto110:
> i don't have problems with actual setup:

Thanks for your reply. I have a RuntimeError if I don't delete sm100 from the setup file.

"RuntimeError: CUDA error: no kernel image is available for execution on the device CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect. For debugging consider passing CUDA_LAUNCH_BLOCKING=1 Compile with TORCH_USE_CUDA_DSA to enable device-side assertions."

PR #795, opened 2025-09-23
build: More fine-grained wheelname for NGC PyTorch

ko3n1g:
The wheels with enabled ABI flags are (from my understanding) meant for the NGC PyTorch containers (since otherwise torch most oftenly comes without that flag).

However, since the PyTorch container picks a torch commit between two releases, it's hard to tell with torch version we should compile the wheel against. I've noticed that mamba 2.2.5 is broken for both NGC PyTorch 25.05 and 25.06.

Intuitively we would fix this by building the wheel inside the container. However due its size, we can't fit this into a public GitHub runner.

This PR enriches wheel names by the custom torch version, so that NGC PyTorch containers don't accidentally pull the wrong version.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -76,6 +76,24 @@ jobs:
       release-version: ${{ needs.setup_release.outputs.release-version }}
       upload-to-release: true
 
+  build_ngc_wheels:
+    name: Build Wheel for NGC PyTorch
+    needs: setup_release
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-20.04]
+        container-image: 
+          - nvcr.io/nvidia/pytorch:25.05-py3
+          - nvcr.io/nvidia/pytorch:25.06-py3
+          - nvcr.io/nvidia/pytorch:25.08-py3
+    uses: ./.github/workflows/_build.yml
+    with:
+      runs-on: ${{ matrix.runs-on }}
+      container-image: ${{ matrix.container-image }}
+      release-version: ${{ needs.setup_release.outputs.release-version }}
+      upload-to-release: true
+
   publish_package:
     name: Publish package
     needs: [build_wheels]


tridao:
I think recent pytorch >= 2.6 have switched to `CXX11_ABI=1`
https://github.com/pytorch/pytorch/releases/tag/v2.6.0

ko3n1g:
> I think recent pytorch >= 2.6 have switched to `CXX11_ABI=1` https://github.com/pytorch/pytorch/releases/tag/v2.6.0

I noticed that the wheels that got compiled with `FORCE_CXX_ABI=True` still break on the latest NGC containers. Do you think that by updating how we set the CXX_ABI flag we can avoid this PR?

tridao:
I actually don't know why `FORCE_CXX_ABI=True` would break NGC. Maybe for torch >= 2.7 we just don't set the flag? Afaik torch >= 2.7 uses CXX11_ABI=1 for all linux wheels

ko3n1g:
Unfortunately also a quick:

```bash
docker run --rm -v $(pwd):/workspace/ -w /workspace/ -it --entrypoint bash --gpus all --shm-size=8g nvcr.io/nvidia/pytorch:25.06-py3     
root@17aa5407f9db:/workspace# wget https://github.com/state-spaces/mamba/releases/download/v2.2.5/mamba_ssm-2.2.5+cu12torch2.8cxx11abiFALSE-cp312-cp312-linux_x86_64.whl
root@17aa5407f9db:/workspace# pip install mamba_ssm-2.2.5+cu12torch2.8cxx11abiFALSE-cp312-cp312-linux_x86_64.whl 
```

Leads to:

```bash
root@17aa5407f9db:/workspace# python 
Python 3.12.3 (main, Feb  4 2025, 14:48:35) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mamba_ssm
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.12/dist-packages/mamba_ssm/__init__.py", line 3, in <module>
    from mamba_ssm.ops.selective_scan_interface import selective_scan_fn, mamba_inner_fn
  File "/usr/local/lib/python3.12/dist-packages/mamba_ssm/ops/selective_scan_interface.py", line 20, in <module>
    import selective_scan_cuda
ImportError: /usr/local/lib/python3.12/dist-packages/selective_scan_cuda.cpython-312-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda9SetDeviceEab
>>> 
```

ko3n1g:
@tridao I think this PR is very close to its final state. The last fixes where mostly around naming the wheel. Do you mind taking a look at the PR already to prefetch some of your thoughts?

PR #810, opened 2025-10-20
add torch 2.8.0 and 2.9.0 for build

okdshin:


benglewis:
Fixes my issue #818

imezx:
hey, what about 2.9.1?

benglewis:
Yes, this would be really great 🙏 

@ko3n1g I would be happy to create a version of this with 2.9.1 and 2.10.0, but I'd like to know that the adjusted pipelines will actually be run 🙏

PR #831, opened 2026-01-02
Fix ROCm 7.0+ compatibility: constexpr WARP_THREADS and lane_id mask for 64-wide wavefronts

AndreasKaratzas:
This PR fixes two issues preventing mamba-ssm from building and running correctly on ROCm 7.0+:

1. **Build failure**: `HIPCUB_WARP_THREADS` is no longer constexpr in ROCm 7.0
2. **Backward pass numerical errors**: Hardcoded 32-thread warp mask breaks 64-thread AMD wavefronts

## Problem 1: Build Failure

Building mamba-ssm on ROCm 7.0+ fails with:
```
error: constexpr variable 'IS_ARCH_WARP' must be initialized by a constant expression
```

### Cause

In ROCm 7.0+, `HIPCUB_WARP_THREADS` expands to `rocprim::warp_size()` which is no longer 
constexpr ([release notes](https://rocm.docs.amd.com/en/latest/about/release-notes.html)).

### Solution

Use `__AMDGCN_WAVEFRONT_SIZE` when available (deprecated but still constexpr in ROCm 7.x), 
with architecture-specific fallbacks for future-proofing.

## Problem 2: Backward Pass Numerical Errors

Backward pass produces incorrect gradients for `seqlen > 128` on AMD GPUs:

| seqlen | Forward | Backward (dA) |
|--------|---------|---------------|
| 128    | ✅ 9.5e-06 | ✅ 0.00 |
| 192    | ✅ 1.1e-05 | ❌ 1247.84 |
| 256    | ✅ 7.6e-06 | ❌ 5572.64 |

### Cause

In `WarpReverseScan` constructor:
```cpp
lane_id(threadIdx.x & 0x1f)  // 0x1f = 31, assumes 32-thread warps
```

This hardcoded mask is incorrect for AMD's 64-thread wavefronts:
- Thread 32: `lane_id = 32 & 31 = 0` (wrong, should be 32)
- Thread 63: `lane_id = 63 & 31 = 31` (wrong, should be 63)

This corrupts warp scan operations for threads 32-63, which only manifests when those 
threads process valid data (seqlen > 128 with kNThreads=64, kNItems=4).

### Solution

Use `WARP_THREADS - 1` as the mask on ROCm to correctly handle both 64-wide (CDNA) 
and 32-wide (RDNA) wavefronts:
```cpp
#ifndef USE_ROCM
    : lane_id(threadIdx.x & 0x1f)              // CUDA: 32-thread warps
#else
    : lane_id(threadIdx.x & (WARP_THREADS - 1)) // ROCm: 64 or 32-thread wavefronts
#endif
```

## Files Changed

- `csrc/selective_scan/reverse_scan.cuh`
- `csrc/selective_scan/reverse_scan_hip.cuh`

## Testing

- ✅ All tests pass `pytest -s -v tests/`, apart from:

```
FAILED tests/ops/triton/test_selective_state_update.py::test_selective_state_update_with_heads[4096-64-2-True-True-itype2] - AssertionError: assert False
```

which fails on both CUDA and ROCm, and is a known failure (it fails before the proposed modifications).

## Version Bump Request

**Please** consider bumping the version to indicate ROCm 7.0+ support, as this enables 
mamba-ssm on the latest AMD software stack.

### DIFF
--- a/csrc/selective_scan/reverse_scan.cuh
+++ b/csrc/selective_scan/reverse_scan.cuh
@@ -101,7 +101,40 @@ struct WarpReverseScan {
     #ifndef USE_ROCM
         #define WARP_THREADS CUB_WARP_THREADS(0)
     #else
-        #define WARP_THREADS HIPCUB_WARP_THREADS
+        // ROCm 7.0+: HIPCUB_WARP_THREADS (rocprim::warp_size()) is no longer constexpr.
+        // We need a compile-time constant for IS_ARCH_WARP below.
+        // See: https://rocm.docs.amd.com/en/latest/about/release-notes.html
+        #if defined(__AMDGCN_WAVEFRONT_SIZE)
+            // Deprecated but still available and constexpr in ROCm 7.x
+            #define WARP_THREADS __AMDGCN_WAVEFRONT_SIZE
+        #elif defined(__gfx942__) || defined(__gfx941__) || defined(__gfx940__)
+            // AMD Instinct MI300 series (CDNA3) - 64-wide wavefronts
+            #define WARP_THREADS 64
+        #elif defined(__gfx90a__)
+            // AMD Instinct MI200 series (CDNA2) - 64-wide wavefronts
+            #define WARP_THREADS 64
+        #elif defined(__gfx908__)
+            // AMD Instinct MI100 (CDNA1) - 64-wide wavefronts
+            #define WARP_THREADS 64
+        #elif defined(__gfx906__) || defined(__gfx900__)
+            // AMD Instinct MI50/MI60 (Vega) - 64-wide wavefronts
+            #define WARP_THREADS 64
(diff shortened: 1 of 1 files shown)


AndreasKaratzas:
cc @tridao 
Hi :) Let me know if you can review this

AndreasKaratzas:
cc @tridao 
Can we get this PR reviewed?

tridao:
Is there no short way to detect WARP_THREADS other than matching on the GPU type?

tridao:
If there's no other way then we can merge as is

AndreasKaratzas:
> If there's no other way then we can merge as is

@tridao Not to my knowledge unfortunately. Sry for the delay, missed your reply 😅

AndreasKaratzas:
Hi @tridao :) Can we merge this PR?

PR #845, opened 2026-02-11
refactor: fold the dt softplus into the kernel on the non-fused Mamba2 path

Mr-Neutr0n:
## What this is

**A refactor, not a bug fix.** The original description claimed `softplus` was applied twice on the non-fused path. That was wrong — see the discussion below. Retitled and rewritten to describe what the change actually does.

## The change

On the non-fused path, `Mamba2Simple.forward` applies the timestep transform in Python before calling the kernel:

```python
dt = F.softplus(dt + self.dt_bias)          # (B, L, nheads)
...
y = mamba_chunk_scan_combined(..., dt, ...)  # dt_bias=None, dt_softplus=False
```

The fused path a few lines above hands `self.dt_bias` straight to `mamba_split_conv1d_scan_combined` and lets the kernel do it. This makes the non-fused branch do the same:

```python
y = mamba_chunk_scan_combined(..., dt_bias=self.dt_bias, dt_softplus=True, **dt_limit_kwargs)
```

Net −1/+2 lines.

## Why it's equivalent

Ordering is preserved. `_chunk_cumsum_fwd` applies `dt_bias` and `softplus` before the `dt_limit` clamp, which is the same order as the current code (Python softplus, then `dt_limit` passed through to the same clamp). `dt_limit_kwargs` is unchanged either way.

## Why bother

- The two branches of the same `forward` currently factor the identical operation two different ways, which is what made the "double softplus" misreading easy in the first place.
- One fewer materialised `(B, L, nheads)` intermediate on the non-fused path.

## What I have not verified

I don't have a GPU to hand, so the Triton path is unexercised locally and the equivalence above is by inspection of `ssd_combined.py`, not by measurement. The backward pass in particular moves from PyTorch autograd through `F.softplus` to `_chunk_cumsum_bwd`'s `ddt_bias` computation. Both should be correct; only one of them is currently exercised from this call site.

If that's not worth the churn on a working path, I'm happy for this to be closed instead — it's a tidiness change, not a fix.

### DIFF
--- a/mamba_ssm/modules/mamba2_simple.py
+++ b/mamba_ssm/modules/mamba2_simple.py
@@ -159,7 +159,6 @@ class Mamba2Simple(nn.Module):
             z, xBC, dt = torch.split(
                 zxbcdt, [self.d_inner, self.d_inner + 2 * self.ngroups * self.d_state, self.nheads], dim=-1
             )
-            dt = F.softplus(dt + self.dt_bias)  # (B, L, nheads)
             assert self.activation in ["silu", "swish"]
 
             # 1D Convolution
@@ -190,6 +189,8 @@ class Mamba2Simple(nn.Module):
                 z=None,
                 seq_idx=seq_idx,
                 initial_states=initial_states,
+                dt_bias=self.dt_bias,
+                dt_softplus=True,
                 **dt_limit_kwargs,
             )
             y = rearrange(y, "b l h p -> b l (h p)")


albertfgu:
Sorry, can you describe the exact code path where the softplus is applied twice?

It seems you're referring to the path starting line 158? Are you claiming the softplus is applied on line 162 and again on line 182 (`mamba_chunk_scan_combined`)? But I don't think the latter applies the softplus by default.

darxradi3nt:
`mamba_chunk_scan_combined` defaults to dt_softplus=False, so it receives the already-transformed dt and applies no further softplus. softplus is applied exactly once.
The proposed change moves the operation inside the Triton kernel via dt_softplus=True, which produces the same result. That may be a cleaner factoring, but there's no double application to fix in the current code.

Suggest re-labeling as a refactor/cleanup rather than a bug fix and close the PR

albertfgu:
Yes, if you relabel this I can merge it

darxradi3nt:
@Mr-Neutr0n?

Mr-Neutr0n:
@darxradi3nt you were right and I owe you a straight answer, four months late. Sorry for the silence, and thanks for pinging.

Checked it properly rather than defending the original claim. `mamba_chunk_scan_combined` defaults to `dt_softplus=False` (`ssd_combined.py:628`), so the call at line 182 applies no softplus of its own and line 162 is the only one. **There is no double application.** My premise was wrong.

What the diff actually does is move the transform from Python into the kernel via `dt_bias=self.dt_bias, dt_softplus=True`. The justification I should have given: the fused branch immediately above already hands `self.dt_bias` to `mamba_split_conv1d_scan_combined` and lets the kernel do it, so the two branches of the same `forward` currently factor the identical operation two different ways. That inconsistency is exactly what made the misreading easy.

Ordering is preserved — `_chunk_cumsum_fwd` applies bias and softplus before the `dt_limit` clamp, same as now.

@albertfgu retitled and rewrote the description as a refactor, per your note. One caveat I'd rather state than have you discover: I have no GPU here, so the Triton path is unexercised locally and the equivalence is by inspection, not measurement. The backward in particular shifts from PyTorch autograd through `F.softplus` to `_chunk_cumsum_bwd`'s `ddt_bias`. If a cosmetic consistency win isn't worth touching a working path unmeasured, closing it is the right call and I won't argue.

Mr-Neutr0n:
Hi @albertfgu - retitled and rewrote the description as a refactor, as you suggested. One caveat I would rather state than have you find: no GPU here, so the Triton path is unexercised and the equivalence is by inspection. If a cosmetic consistency win is not worth touching a working path unmeasured, closing is completely fine and I will not argue.

PR #856, opened 2026-03-13
Respect TORCH_CUDA_ARCH_LIST to speed up builds

### LINKED ISSUE #625: TORCH_CUDA_ARCH_LIST support
Can we support arch list specification with this env as documented in
https://pytorch.org/docs/stable/cpp_extension.html ?

d3banjan:
Fixes #625

## Summary
- When `TORCH_CUDA_ARCH_LIST` is set, parse it and generate only the requested `-gencode` flags instead of hardcoding all supported architectures
- When unset, behavior is completely unchanged (existing hardcoded flags remain as fallback)
- This is the standard PyTorch convention already used by [flash-attention](https://github.com/Dao-AILab/flash-attention), [xformers](https://github.com/facebookresearch/xformers), and PyTorch's own `cpp_extension.py`

## Motivation
Building from source currently compiles for all supported GPU architectures regardless of the target hardware. For users targeting a single architecture (e.g. `TORCH_CUDA_ARCH_LIST="8.6"`), this makes builds ~5–7x slower than necessary.

## Test plan
- [ ] `pip install -e .` with `TORCH_CUDA_ARCH_LIST` **unset** — verify all existing gencode flags are emitted (unchanged behavior)
- [ ] `TORCH_CUDA_ARCH_LIST="8.6" pip install -e .` — verify only `-gencode arch=compute_86,code=sm_86` appears in nvcc output
- [ ] `TORCH_CUDA_ARCH_LIST="7.5;8.0" pip install -e .` — verify both architectures are emitted
- [ ] `TORCH_CUDA_ARCH_LIST="8.6+PTX" pip install -e .` — verify PTX suffix is stripped and `compute_86` is used

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -172,27 +172,42 @@ if not SKIP_CUDA_BUILD:
                     "Note: make sure nvcc has a supported version by running nvcc -V."
                 )
 
-        cc_flag.append("-gencode")
-        cc_flag.append("arch=compute_75,code=sm_75")
-        cc_flag.append("-gencode")
-        cc_flag.append("arch=compute_80,code=sm_80")
-        cc_flag.append("-gencode")
-        cc_flag.append("arch=compute_87,code=sm_87")
-        if bare_metal_version >= Version("11.8"):
+        # Allow overriding CUDA architectures via TORCH_CUDA_ARCH_LIST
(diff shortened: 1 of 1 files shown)


kevalmorabia97:
+1. Would be great to have this merged

PR #863, opened 2026-03-20
fix: add missing dependencies for Mamba3 (tilelang, nvidia-cutlass-dsl)

Ollie-spoon:
Currently, attempting to import Mamba3 on a fresh installation results in `ModuleNotFoundError` crashes because tilelang and CuTe are not listed in the build dependencies. This PR adds them to install_requires in setup.py so the package builds and imports cleanly out of the box. 

This PR also updates the quack-kernels requirement to be "quack-kernels==0.3.4" rather than "quack-kernels==0.3.1".


tridao:
the quack dependency should already have nvidia-cutlass-dsl as a dependecy

PR #878, opened 2026-03-24
Mamba-3: imports fail silently

berlinchen7:
Ensure a mamba-2-only environment use mamba-2 without extra mamba-3 packages.


kevinli573:
lgtm, merged!

PR #890, opened 2026-03-30
Feat: varlen support for Mamba-3

berlinchen7:
Add variable-length (varlen) support for Mamba-3 MIMO.

Mamba-3 SISO already supports varlen, and this change extends the same capability to MIMO, enabling consistent handling of variable-length inputs across both variants.

### DIFF
--- a/mamba_ssm/modules/mamba3.py
+++ b/mamba_ssm/modules/mamba3.py
@@ -9,13 +9,20 @@ import torch.nn.functional as F
 
 from mamba_ssm.ops.triton.layernorm_gated import RMSNorm as RMSNormGated
 
-from mamba_ssm.ops.tilelang.mamba3.mamba3_mimo import mamba3_mimo as mamba3_mimo_combined
+try:
+    from mamba_ssm.ops.tilelang.mamba3.mamba3_mimo import mamba3_mimo as mamba3_mimo_combined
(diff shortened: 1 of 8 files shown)


skaae:
Can you share what environment you you used to build the mimo kernels? 

- Cuda version, 
- gpu model, 
- pytorch version 
- gcc compiler

I have tried using pytorch 2.11 and cuda 13.0 and n RTX 6000 Pro Blackwell. The code compiles but fails at runtime with 

```
tvm::runtime::detail::LogFatal::Entry::Finalize()
tvm.error.InternalError: Failed to set the allowed dynamic shared memory size to 146080
```
Are you compiling on hopper?

from the setup: https://github.com/state-spaces/mamba/blob/31f3d7baba69d0ccad1635ace1e477367899e408/setup.py#L187-L201

it looks like you hopper would only compile using cuda 11.8 since hobber is sm_90?

berlinchen7:
From the error message, it appears there may be an over-allocation of shared memory. As a quick check, you could try reducing the chunk size by half, which should lower the shared memory footprint of the MIMO kernels.

To answer your question, the environment details are as follows:
- CUDA version: 13.1
- GPU: H100
- PyTorch version: 2.9.1
- GCC version: 11.5.0

PR #896, opened 2026-04-06
Sync quack-kernels to 0.3.4 in pyproject.toml

Chessing234:
## Summary
- `pyproject.toml` pins `quack-kernels==0.3.1` while `setup.py` pins `quack-kernels==0.3.4`
- This mismatch causes PEP 517 builds (`pip install .`) to install the older version
- Sync `pyproject.toml` to match `setup.py` at `0.3.4`

## Test plan
- [x] Verified `setup.py` line 403 specifies `quack-kernels==0.3.4`
- [x] Single version bump, no other changes

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ classifiers = [
 dependencies = [
     "torch",
     "tilelang==0.1.8",
-    "quack-kernels==0.3.1",
+    "quack-kernels==0.3.4",
     "triton>=3.5.0",
     "ninja",
     "einops",


tridao:
We should just set quack-kernels >= 0.3.4

tridao:
@caitWW @aakashlahoti is that the right version of quack?

caitWW:
let me check I think 0.3.4 should also work but when we wrote the code it was still 0.3.1 also w older version of nvidia-cutlass

caitWW:
I've bumped quack-kernels to >= 0.3.4 and updated cute dsl code to newer version. @Chessing234 thanks for bringing this up!

Chessing234:
Thanks @tridao and @caitWW for the quick response and for bumping it to >= 0.3.4! Appreciate you taking care of it.

PR #901, opened 2026-04-06
Fix Mamba2.step() handling of D when D_has_hdim=True

### LINKED ISSUE #887: `Mamba2.step()` handles `D` incorrectly when `D_has_dim=True`
Hi, I think there's a bug in `Mamba2.step()` when we set `D_has_dim = True`. 

In `__init__`, `self.D` is initialized as
```
self.D = nn.Parameter(torch.ones(self.d_ssm if self.D_has_hdim else self.nheads, device=device))
```
Thus, when `D_has_hdim=True`, `self.D` has shape `(d_ssm,) = (nheads * headdim,)`.

I believe the forward path appear to handle this correctly by reshaping `D` as `(h, p)`
```
So when D_has_hdim=True, self.D has shape (d_ssm,) = (nheads * headdim,).
```
But I don't think `step()` makes the same distinction because for [this line](https://github.com/state-spaces/mamba/blob/main/mamba_ssm/modules/mamba2.py#L327), it implicitly assumes that `self.D` is per head, with shape `(nheads,)`, and not per head-dim: `(nheads * headdim,)`. 

So for `D_has_hdim = True`, `step()` seems inconsistent with `forward()`, as `forward()` reshapes `self.D` from `(h * p,) -> (h,p)`, whereas `step()` instead treats it as it were `(h,)`, as mentioned earlier

I can contribute a PR to fix this

Chessing234:
## Summary

- Fixes `Mamba2.step()` to correctly handle `self.D` when `D_has_hdim=True`, where `D` has shape `(nheads * headdim,)` instead of `(nheads,)`
- Both code paths in `step()` (with and without `selective_state_update`) now use `rearrange("(h p) -> h p")` when `D_has_hdim=True`, consistent with `forward()`
- Without this fix, autoregressive decoding produces incorrect outputs when `D_has_hdim=True` because `D` is misinterpreted as per-head rather than per-head-dim

Fixes #887

## Test plan

- [ ] Verify `step()` output matches `forward()` output for a single token when `D_has_hdim=True`
- [ ] Confirm no regression when `D_has_hdim=False` (default behavior unchanged)
- [ ] Test both code paths: with `selective_state_update` available and with the fallback path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mamba_ssm/modules/mamba2.py
+++ b/mamba_ssm/modules/mamba2.py
@@ -316,7 +316,10 @@ class Mamba2(nn.Module, PyTorchModelHubMixin):
             dBx = torch.einsum("bh,bn,bhp->bhpn", dt, B, x)
             ssm_state.copy_(ssm_state * rearrange(dA, "b h -> b h 1 1") + dBx)
             y = torch.einsum("bhpn,bn->bhp", ssm_state.to(dtype), C)
-            y = y + rearrange(self.D.to(dtype), "h -> h 1") * x
+            if self.D_has_hdim:
+                y = y + rearrange(self.D.to(dtype), "(h p) -> h p", p=self.headdim) * x
+            else:
+                y = y + rearrange(self.D.to(dtype), "h -> h 1") * x
             y = rearrange(y, "b h p -> b (h p)")
             if not self.rmsnorm:
                 y = y * self.act(z)  # (B D)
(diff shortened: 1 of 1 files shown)


Chessing234:
Closing in favor of #903 (duplicate)

Issue #888, opened 2026-03-28
Mamba2 `step()` works silently but when D_has_hdim=True and selective_state_update=None but is


PuR3Luck:
There seems to be an issue on https://github.com/state-spaces/mamba/blob/main/mamba_ssm/modules/mamba2.py#L319

When `D_has_hdim=True`, `self.D` has shape `(d_ssm)  = (n_heads * headdim,)`

So the the rearrange is unsqueezing the last dimension `(n_heads * headdim,) -> (n_heads * headdim, 1)`.
However the shape of `x` and `y` is `(b h p)` which cannot be broadcasted to the shape of the rearranged D `(n_heads * headdim, 1)`

This behaviour is inconsistent with the `forward()` method 

If needed, I can contribute a PR for this 

Related issue: https://github.com/state-spaces/mamba/issues/887

ccs1112:
Same root cause as #887 — fix and test at #893.

PR #903, opened 2026-04-06
Fix Mamba2.step() D handling when D_has_hdim=True

### LINKED ISSUE #887: `Mamba2.step()` handles `D` incorrectly when `D_has_dim=True`

Chessing234:
## Summary
- When `D_has_hdim=True`, `self.D` has shape `(d_ssm,)` = `(nheads * headdim,)`, but `step()` was treating it as shape `(nheads,)` in both the fallback path and the `selective_state_update` path.
- Added conditional reshaping (`rearrange(self.D, "(h p) -> h p", p=self.headdim)`) to both code paths in `step()`, consistent with how `forward()` already handles `D_has_hdim=True`.

## Details
In `step()`, there are two code paths that use `self.D`:

1. **Fallback path** (when `selective_state_update is None`): was doing `rearrange(self.D, "h -> h 1")` which assumes D has shape `(nheads,)`. Now conditionally uses `rearrange(self.D, "(h p) -> h p", p=self.headdim)` when `D_has_hdim=True`.

2. **`selective_state_update` path**: was doing `repeat(self.D, "h -> h p", p=self.headdim)` which also assumes D has shape `(nheads,)`. Now conditionally uses `rearrange` instead of `repeat` when `D_has_hdim=True`.

Both `forward()` code paths (lines 191 and 251) already handled this correctly via:
```python
D=rearrange(self.D, "(h p) -> h p", p=self.headdim) if self.D_has_hdim else self.D
```

Fixes #887

## Test plan
- [ ] Instantiate `Mamba2` with `D_has_hdim=True` and verify `step()` produces correct output matching `forward()` for single-token inputs
- [ ] Verify shapes are consistent: `self.D` shape `(d_ssm,)` is reshaped to `(nheads, headdim)` before use in both step paths
- [ ] Verify no regression when `D_has_hdim=False` (default behavior unchanged)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mamba_ssm/modules/mamba2.py
+++ b/mamba_ssm/modules/mamba2.py
@@ -316,7 +316,10 @@ class Mamba2(nn.Module, PyTorchModelHubMixin):
             dBx = torch.einsum("bh,bn,bhp->bhpn", dt, B, x)
             ssm_state.copy_(ssm_state * rearrange(dA, "b h -> b h 1 1") + dBx)
             y = torch.einsum("bhpn,bn->bhp", ssm_state.to(dtype), C)
-            y = y + rearrange(self.D.to(dtype), "h -> h 1") * x
+            if self.D_has_hdim:
+                y = y + rearrange(self.D.to(dtype), "(h p) -> h p", p=self.headdim) * x
+            else:
+                y = y + rearrange(self.D.to(dtype), "h -> h 1") * x
             y = rearrange(y, "b h p -> b (h p)")
             if not self.rmsnorm:
                 y = y * self.act(z)  # (B D)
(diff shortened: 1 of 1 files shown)


ccs668899:
I opened #893 on April 2 to address the same bug and also fix #888. It includes a parametrized test `test_mamba2_step_forward_consistency` covering both `D_has_hdim` cases with non-uniform D values to ensure the bug isn't masked.

Chessing234:
Closing as #893 (opened earlier by @ccs668899) already covers both this fix and #888, plus a parametrized `test_mamba2_step_forward_consistency` that catches the bug. Deferring to that PR.

PR #922, opened 2026-04-16
Update angle normalization using remainder function

sukjunhwang:
`angle` might grow unboundedly if sequence length gets very large.
Therefore, using remainder for angle normalization.


tridao:
cc @aakashlahoti @berlinchen7

PR #937, opened 2026-05-03
Set `batch`, `seqlen`, `ngroups`, and `nheads` as dynamic arguments for MIMO kernels.

aakashlahoti:
To avoid recompiling MIMO kernels when `batch`, `seqlen`, `ngroups`, or `nheads` changes, pass them as `T.dynamic` input arguments during kernel compilation.


aakashlahoti:
+@berlinchen7

PR #960, opened 2026-05-30
Catch InvalidVersion in setup.py to fall back to source build

### LINKED ISSUE #947: mamba-ssm setup.py fails with InvalidVersion for vendor PyTorch/CUDA version strings instead of falling back to source build
## Summary

Installing `mamba-ssm` can fail during wheel build when `setup.py:get_wheel_url()` parses vendor-specific PyTorch/CUDA version metadata with `packaging.version.parse`. If the environment contains a non-PEP-440 vendor build string, `get_wheel_url()` raises `packaging.version.InvalidVersion` and aborts the install instead of falling back to a source build.

This is reproducible in vendor PyTorch containers where CUDA/PyTorch build metadata may be exported as values like `gpgpu.<build-id>`.

## Observed failure

Package:

- `mamba-ssm==2.3.2.post1`
- Python 3.12
- Linux x86_64
- Vendor/internal PyTorch prerelease container

During installation as a transitive dependency:

```text
Building mamba-ssm==2.3.2.post1
× Failed to build `mamba-ssm==2.3.2.post1`
...
File "<string>", line 340, in run
File "<string>", line 315, in get_wheel_url
File ".../packaging/version.py", line 103, in parse
  return Version(version)
File ".../packaging/version.py", line 361, in __init__
  raise InvalidVersion(f"Invalid version: {version!r}")
packaging.version.InvalidVersion: Invalid version: 'gpgpu.<build-id>'

Chessing234:
## Bug

`get_wheel_url()` in `setup.py` calls `packaging.version.parse()` on `torch.__version__` and `torch.version.cuda`. In vendor PyTorch containers these can be non-PEP 440 strings (e.g. `gpgpu.<build-id>`), causing `parse()` to raise `InvalidVersion`.

Because `CachedWheelsCommand.run()` has no handler for this exception, the install aborts instead of falling back to a source build.

## Root cause

`get_wheel_url()` (line 281/290) passes raw version strings to `packaging.version.parse()` without guarding against `InvalidVersion`. The existing `try/except` in `run()` only catches `urllib.error.HTTPError` from the download step — it never reaches that point.

## Fix

Wrap the `get_wheel_url()` call with `try/except InvalidVersion` so the build falls back to `super().run()` (source compilation) when the version string cannot be parsed. This matches the existing fallback behavior for missing prebuilt wheels.

Fixes #947

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import os
 import re
 import ast
 from pathlib import Path
-from packaging.version import parse, Version
+from packaging.version import parse, InvalidVersion, Version
 import platform
 import shutil
 
@@ -337,7 +337,11 @@ class CachedWheelsCommand(_bdist_wheel):
         if FORCE_BUILD:
             return super().run()
 
-        wheel_url, wheel_filename = get_wheel_url()
+        try:
+            wheel_url, wheel_filename = get_wheel_url()
+        except InvalidVersion:
+            print("Could not determine wheel URL (non-PEP 440 version). Building from source...")
+            return super().run()
         print("Guessing wheel URL: ", wheel_url)
         try:
             urllib.request.urlretrieve(wheel_url, wheel_filename)


Chessing234:
Closing as duplicate of #961.

PR #976, opened 2026-06-15
Add support for input states to Mamba-3 forward pass

aakashlahoti:
Enable Mamba-3 SISO and MIMO codepaths to accept input states for forward pass/ prefill.


Chessing234:
input states support is useful. please document the expected state layout (dtype/device/batch) in the docstring.

PR #978, opened 2026-06-20
Fix Mamba-3 decode path: forward() to step() shape handling

sanowl:
Mamba3.forward() routes to step() during autoregressive decode (seqlen_offset > 0),
passing the (batch, 1, dim) tensor it receives. step() expects (batch, dim):
_preprocess uses 2-axis einops patterns such as rearrange(x, "b (h p) -> b h p").
Decoding through the module crashes at the first generated token. Even without the
crash, the previous `return out` returns 2D while the prefill path returns 3D.
Mamba2.step() handles this by squeezing the input and unsqueezing the output;
Mamba3 did neither.

 the Fix
- Squeeze the singleton seqlen before step(), restore it after, with an
  assert seqlen == 1 guard (single-token decode, matching Mamba2).
- Add test_forward_decode_dispatch_matches_step: prefill via forward, decode
  through the seqlen_offset > 0 branch, and check parity against the direct-step
  reference (SISO and MIMO, out-norm on and off).

I have not run this. No NVIDIA GPU available locally (the CUTE step kernel requires
H100). The change is verified by code review and matches the existing Mamba-2
convention; the added test should validate it on CUDA / CI.

### DIFF
--- a/mamba_ssm/modules/mamba3.py
+++ b/mamba_ssm/modules/mamba3.py
@@ -169,8 +169,13 @@ class Mamba3(nn.Module):
             inference_batch = cu_seqlens.shape[0] - 1 if cu_seqlens is not None else batch
             angle_dt_state, ssm_state, k_state, v_state = self._get_states_from_cache(inference_params, inference_batch)
             if inference_params.seqlen_offset > 0:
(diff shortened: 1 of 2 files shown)


Chessing234:
decode path shape fix looks right. does step() still match forward() numerics on a short prompt+one token case?

PR #988, opened 2026-07-09
Audit follow-up: fix the same int32 overflow pattern in other Triton kernels

chipsncheese-ie:
**Stacked on #987** — this branch is based on that one's branch (`fix/int32-overflow-chunk-cumsum`), so the diff shown here currently includes those changes too until #987 merges. Once #987 lands this diff will shrink to just the commits below.

## Summary

Follow-up audit after #987 fixed the int32 pointer-arithmetic overflow in `_chunk_cumsum_fwd_kernel`/`_chunk_cumsum_bwd_kernel`. The same unguarded `pid_* * chunk_size * stride_*_seqlen`-style chained multiplication, with an uncast 32-bit leading operand, turned out to exist in essentially every other `pid_bc`/`pid_c`/`pid_ch`-indexed kernel in the Triton ops:

- `ssd_chunk_state.py` — the remaining kernels: `_chunk_state_fwd_kernel`, `_chunk_state_bwd_dx_kernel`, `_chunk_state_bwd_db_kernel`, `_chunk_state_bwd_ddAcs_stable_kernel`, `_chunk_state_varlen_kernel`
- `ssd_chunk_scan.py` — all eleven `pid_bc`-indexed kernels
- `ssd_combined.py` — the fused kernel
- `ssd_bmm.py` — both kernels

Same fix in each case: cast the leading program-id-derived index to `tl.int64` immediately after it's obtained, mirroring the fix in #987.

**A related but distinct instance of the same overflow class** turned up in `layer_norm.py`, `layernorm_gated.py`, and `k_activations.py`: these kernels compute `row * stride_x_row` where `row = tl.program_id(0)` ranges over all flattened rows (`batch*seqlen`), and `stride_x_row` is the row's width (hidden dim). Unlike the chunk-indexed kernels, this doesn't need a non-contiguous view to overflow — an ordinarily *contiguous* `(M, N)` input is enough once `M * N` exceeds `2**31 - 1`, since `row` alone ranges up to `M-1`. Same fix: cast `row`/`row_block_id` to `tl.int64` right after `tl.program_id()`.

Every finding here was confirmed as a **real, reproducible crash** (not just a theoretical read of the code) before being fixed:

- `mamba_chunk_scan_combined`, called with `x` sliced non-contiguously out of a wide parent tensor (same shape as the original #987 bug report), crashed with the identical `Triton Error [CUDA]: an illegal memory access was encountered` pre-fix — tracing into `_chunk_state_fwd_kernel`, a kernel #987 didn't touch. Passes post-fix.
- `layernorm_fn` (RMSNorm path) on a contiguous `(115_866, 18_560)` input crashed identically pre-fix, passes post-fix.

## Not fixed here, flagged for follow-up

- `ssd_state_passing.py` has `(nchunks - 1) * stride_*_chunk` with an uncast `nchunks`, structurally the same class of risk, but `nchunks` isn't program-id-derived and `stride_*_chunk` is a state-tensor stride (`nheads * headdim`-ish) rather than a wide/sliceable seqlen stride — I couldn't construct a concrete overflow scenario for it, so I'm flagging it rather than speculatively patching it.
- `selective_state_update.py` only multiplies `pid_b`/`pid_h` against batch/head strides for single-token inference state updates — lower risk, not touched.
- `mamba_ssm/ops/triton/mamba3/` (a separate, newer subsystem) uses `pid_chunk * stride_*_chunk` as a single multiplication against a chunk-dim stride, not a chained product through a wide seqlen stride — a different shape of risk that I haven't audited in depth.

## Test plan

- [x] Added `test_mamba_chunk_scan_combined_noncontiguous_wide_view_no_overflow` in `tests/ops/triton/test_ssd.py` — exercises the `ssd_chunk_state.py`, `ssd_chunk_scan.py`, `ssd_combined.py`, and `ssd_bmm.py` fixes together via the full `mamba_chunk_scan_combined` path. Verified it fails with the illegal-memory-access error on the unpatched kernels and passes with the fix.
- [x] Added `test_layer_norm_gated_large_row_count_no_overflow` in `tests/ops/triton/test_layernorm_gated.py`. Verified it fails identically pre-fix and passes post-fix.

### DIFF
--- a/mamba_ssm/ops/triton/k_activations.py
+++ b/mamba_ssm/ops/triton/k_activations.py
@@ -31,7 +31,9 @@ def _swiglu_fwd_kernel(
     BLOCK_N: tl.constexpr,
 ):
     # Map the program id to the row of X and Y it should compute.
-    row = tl.program_id(0)
+    # if row * stride_x_row is large, may overflow int32, so use 64 bit
+    # https://github.com/triton-lang/triton/issues/1058
+    row = tl.program_id(0).to(tl.int64)
     start_col = tl.program_id(1) * BLOCK_N
     X += row * stride_x_row
     Y += row * stride_y_row
@@ -93,7 +95,9 @@ def _swiglu_bwd_kernel(
     RECOMPUTE_OUTPUT: tl.constexpr,
 ):
     # Map the program id to the row of X and Y it should compute.
-    row = tl.program_id(0)
+    # if row * stride_x_row is large, may overflow int32, so use 64 bit
+    # https://github.com/triton-lang/triton/issues/1058
+    row = tl.program_id(0).to(tl.int64)
     start_col = tl.program_id(1) * BLOCK_N
     X += row * stride_x_row
     Y += row * stride_y_row
--- a/mamba_ssm/ops/triton/layer_norm.py
+++ b/mamba_ssm/ops/triton/layer_norm.py
@@ -217,7 +217,9 @@ def _layer_norm_fwd_1pass_kernel(
     HAS_B1: tl.constexpr,
 ):
     # Map the program id to the row of X and Y it should compute.
-    row = tl.program_id(0)
+    # if row * stride_x_row is large, may overflow int32, so use 64 bit
+    # https://github.com/triton-lang/triton/issues/1058
+    row = tl.program_id(0).to(tl.int64)
     X += row * stride_x_row
     Y += row * stride_y_row
     if HAS_RESIDUAL:
@@ -479,7 +481,9 @@ def _layer_norm_bwd_kernel(
     RECOMPUTE_OUTPUT: tl.constexpr,
 ):
     # Map the program id to the elements of X, DX, and DY it should compute.
-    row_block_id = tl.program_id(0)
+    # if row_start * stride_x_row is large, may overflow int32, so use 64 bit
+    # https://github.com/triton-lang/triton/issues/1058
+    row_block_id = tl.program_id(0).to(tl.int64)
     row_start = row_block_id * rows_per_program
     # Do not early exit if row_start >= M, because we need to write DW and DB
     cols = tl.arange(0, BLOCK_N)
--- a/mamba_ssm/ops/triton/layernorm_gated.py
+++ b/mamba_ssm/ops/triton/layernorm_gated.py
@@ -63,7 +63,9 @@ def _layer_norm_fwd_1pass_kernel(
     IS_RMS_NORM: tl.constexpr,
 ):
     # Map the program id to the row of X and Y it should compute.
-    row = tl.program_id(0)
+    # if row * stride_x_row is large, may overflow int32, so use 64 bit
+    # https://github.com/triton-lang/triton/issues/1058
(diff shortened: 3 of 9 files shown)


KyleMylonakisProtopia:
Any update on landing this?

KyleMylonakisProtopia:
Thanks @chipsncheese-ie for the thorough suggestion. 

One gap survives in the two `_chunk_cumsum_*` kernels this diff carries over from #987: the **batch axis (`pid_b`, axis 0) is still left 32-bit**. Only `pid_c` was cast:

```python
# ssd_chunk_state.py — _chunk_cumsum_fwd_kernel (and identically _chunk_cumsum_bwd_kernel)
pid_b = tl.program_id(axis=0)                 # <-- still int32
pid_c = tl.program_id(axis=1).to(tl.int64)
pid_h = tl.program_id(axis=2)
dt_ptr += pid_b * stride_dt_batch + pid_c * chunk_size * stride_dt_seqlen
```

`pid_c * chunk_size * stride_dt_seqlen` is now safe, but `pid_b * stride_dt_batch` is still evaluated in int32. When `dt` is a non-contiguous view into a wide parent tensor — the HF-transformers NemotronH case, where `time_step` is a `torch.split` of the fused `in_proj` output, so `stride_dt_batch = seqlen * d_in_proj` — this term overflows for `batch >= 2`, at sequence lengths well short of the single-batch threshold.

**Concrete, reproduced crash** (Nemotron-H Ultra): `batch=4`, `seqlen=40960`, `d_in_proj=35072` → `stride_dt_batch = 40960 * 35072 = 1,436,549,120`. `pid_b=2` gives `2.87e9 > 2**31-1`, so batch elements 2 and 3 wrap to a negative element offset → `Triton Error [CUDA]: an illegal memory access was encountered` inside `_chunk_cumsum_fwd_kernel` (and its bwd twin on the backward pass). It showed up during activation-checkpoint recomputation in `NemotronHMamba2Mixer`.

Both of the new regression tests slice out of a single-batch parent (`(1, seqlen, parent_width)`), so `pid_b` is always 0 and the batch-axis term is never exercised — which is why this slips through.

Same one-liner fixes both kernels:

```python
pid_b = tl.program_id(axis=0).to(tl.int64)
```

(only these two kernels need it — the others fold batch into the already-cast `pid_bc`). A `batch >= 2` + wide-view case in the regression test would lock it in. Since the two `_chunk_cumsum_*` kernels landed in #987, the cast arguably belongs there, but flagging it here on the active audit PR.

Chessing234:
good follow-up. can you list which kernels you audited so reviewers don't have to rediscover the same pattern?

KyleMylonakisProtopia:
> good follow-up. can you list which kernels you audited so reviewers don't have to rediscover the same pattern?

I was planning on opening up a PR with all the spots we found. The issue is that there could be performance losses due to `torch.contiguous` calls. Basically any time we have view's of large tensors, and we have 32 bit indexing, this can crop up.

chipsncheese-ie:
Closing this in favor of a new PR.

Issue #986, opened 2026-07-08
apache-tvm-ffi<=0.1.12 (after #984) is incompatible with pinned tilelang==0.1.8 — measured working ceiling is 0.1.9


saromleang:
### Summary

  After #984 raised the bound to `apache-tvm-ffi<=0.1.12` (`pyproject.toml:21`, `setup.py:403`), a fresh source install resolves tvm-ffi versions that are
  incompatible with the pinned `tilelang==0.1.8`. We measured the full version matrix on a clean environment; only **0.1.9 and below** work.

  ### Version matrix (measured)

  | apache-tvm-ffi | Result |
  |---|---|
  | 0.1.12 (current resolver pick) | `import mamba_ssm` crashes: `AttributeError: attribute '__dict__' of 'type' objects is not writable` in
  `tvm_ffi/registry.py` (`_add_class_attrs`). The import chain `mamba_ssm/__init__ → modules.mamba3 → ops.tilelang… → import tilelang` is unconditional, so the
  whole package is unimportable. |
  | 0.1.11, 0.1.10 | Import OK, but **every** tilelang kernel compile fails: `AttributeError: '_NestedLoopCheckVisitor' object has no attribute '_inst'`
  (`tilelang/analysis/nested_loop_checker.py`, via `tvm/runtime/support.py`) — same failure as #907. All 90+ `tests/ops/tilelang/test_mamba3_mimo.py` tests
  fail. |
  | **0.1.9, 0.1.8.post2** | **Works.** All tilelang forward-kernel tests pass on our box (remaining failures are an unrelated sm_121 shared-memory ceiling,
  see note below). |

  tilelang 0.1.8 itself declares `apache-tvm-ffi>=0.1.2,~=0.1.0`, so the resolver happily picks the broken versions unless mamba constrains them.

  ### Suggestion

  While `tilelang==0.1.8` is pinned, the compatible constraint appears to be `apache-tvm-ffi<0.1.10`, not `<=0.1.12`. #984's motivation was flash-attn-4
  wanting `>=0.1.12` — with tilelang 0.1.8 pinned those two requirements look mutually exclusive, so this may need a tilelang bump rather than a bound tweak.

  Related catch-22 we hit while validating: `nvidia-cutlass-dsl 4.6.0` (the `tests/ops/cute` path) calls `make_kwargs_wrapper(map_dataclass_to_tuple=…)`, which
[...]

PR #1001, opened 2026-07-29
Fix: support TileLang 0.1.9

benglewis:
## Summary

- update the exact TileLang pin from 0.1.8 to 0.1.9
- cap apache-tvm-ffi below 0.1.10, matching the working compatibility ceiling identified in #986

This allows Mamba to resolve alongside vLLM releases that require TileLang 0.1.9.

## Validation

- verified pyproject.toml and setup.py carry the same dependency pair
- `git diff --check`

A CUDA Mamba/MIMO smoke test is still required before publishing a rebuilt wheel.


tridao:
cc @berlinchen7 
I did try 0.1.9 at some point very briefly and saw some perf regression on some other kernel, so we should check that the MIMO kernels don't regress after updating tilelang.

Chessing234:
looks reasonable. maybe pin an upper bound too so the next tilelang break doesn't silently install?

benglewis:
@Chessing234 I have added the pin on the upper bound instead of the fixed version. @tridao How can I test the MIMO kernels? What GPU or similar would I need to test and how?

bvolpato (commented):
Could we allow `apache-tvm-ffi==0.1.10` here? vLLM v0.26 pins it alongside `tilelang==0.1.9`, so the current `<0.1.10` bound still makes that combination unsatisfiable. I tested TileLang 0.1.9 with tvm-ffi 0.1.10 and sampled MIMO forward/backward cases passed.

Could we also run a broader MIMO perf pass before raising the floor? On SM120 I measured one N64 forward shape about 5.4% slower with TileLang 0.1.9, 0.6369 ms versus 0.6042 ms across three medians, while the smaller N16 case was flat. This looks shape-specific, but seems worth checking.

PR #1003, opened 2026-07-30
[`Triton`] Fix triton launch grids under multi gpu scenarios

vasqu:
It's essentially the same underlying issue as https://github.com/Dao-AILab/flash-attention/issues/523

Triton can have issues to select the correct underlying device


kingtweak69:
!

Chessing234:
nice catch on the launch grid — can you add a tiny multi-gpu smoke test or at least a comment with the failing shape so this doesn't regress quietly?

vasqu:
cc @tridao whenever you have time

Seems like nemotron h tried to fix it in a wrong way so I'd rather have this upstream (they pinned the complete cuda stream)

Issue #999, opened 2026-07-23
Publish wheels for torch 2.11 / 2.12 / 2.13 — the PyPI matrix has been pinned at 2.10 since February


fabianbergermann:
## TL;DR: Would you be open to publish wheels for recent torch versions? 
Many thanks for the amazing work! 
I'm happy to hand in the (mini) PR for the above ask, if you  prefer me to hand in a PR instead of "just" an issue.

Note: Claude-assisted context and concrete steps below:

## Summary

`.github/workflows/publish.yaml` hardcodes the PyPI torch build matrix:

```yaml
torch-version: ["2.6.0", "2.7.1", "2.8.0", "2.9.1", "2.10.0"]
```

torch has shipped three minor releases past that since:

| torch | released on PyPI |
| --- | --- |
| 2.10.0 | 2026-02-10 |
| 2.11.0 | 2026-03-23 |
| 2.12.0 | 2026-05-13 |
| 2.12.1 | 2026-06-17 |
| 2.13.0 | 2026-07-08 |

The most recent release, `v2.3.2.post1` (2026-05-09), still built against 2.10.0 as
its ceiling — torch 2.11.0 was already seven weeks old at that point. So this isn't
"a release hasn't been cut lately"; the matrix itself is the thing that has gone
stale, and cutting a release without touching it wouldn't change anything.

## Why it matters downstream

PyPI carries only sdists for `mamba-ssm` and `causal-conv1d`, so anyone whose
resolver lands on torch >= 2.11 silently falls off the wheel path and compiles the
CUDA kernels from source. On our side that is **15–35 minutes on a 48-core box**, and
considerably worse on the ~9-vCPU instances that are usually what's actually
available to rent. It is also a *silent* fallback — no warning that a wheel was
missed, the only symptom is half an hour of apparent hang.

Because a plain `torch>=2.x` constraint floats to the newest release, this is the
default outcome for anyone not explicitly pinning. We ended up adding a hard
ceiling to work around it:

```toml
# pyproject.toml
"torch>=2.9,<2.11",   # upper bound only because prebuilt mamba-ssm wheels stop at 2.10

[tool.uv.sources]
mamba-ssm = [{ url = "https://github.com/state-spaces/mamba/releases/download/v2.3.2.post1/mamba_ssm-2.3.2.post1+cu12torch2.10cxx11abiTRUE-cp312-cp312-linux_x86_64.whl" }]
[...]

PR #1005, opened 2026-07-31
ci: publish wheels for torch 2.11, 2.12.1, and 2.13.0

Chessing234:
## Summary
- Extend `.github/workflows/publish.yaml` to build release wheels for torch 2.11.0, 2.12.1, and 2.13.0 with CUDA 11.8 excludes and CXX11_ABI excludes (#999).
- Document in README how prebuilt wheel coverage tracks the release matrix.
- Add maintainer comment linking the matrix to `setup.py` wheel URL tags.

## Test plan
- [ ] Requires a tagged release to exercise the expanded matrix in CI
- [x] YAML/matrix diff reviewed against existing exclude rules

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -108,6 +108,9 @@ jobs:
               '2.8': [126, 128, 129], \
               '2.9': [126, 128, 130], \
               '2.10': [126, 128, 130], \
+              '2.11': [126, 128, 129, 130], \
+              '2.12': [126, 128, 129, 130], \
+              '2.13': [126, 128, 129, 130], \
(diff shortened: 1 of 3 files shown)


bvolpato (commented):
Could we also add 2.11, 2.12, and 2.13 to the `available` map in `.github/workflows/_build.yml`? I reproduced each new job failing with `KeyError` before `pip install` because that map currently stops at 2.10. cu129 and cu130 inputs are available for both architectures, so adding those entries should unblock this matrix.

Chessing234:
added 2.11/2.12/2.13 to the available map in _build.yml with cu126/128/129/130.

PR #1031, opened 2026-08-31
disregard p2

ajf-codes:
Bumps the `Jimver/cuda-toolkit` action from `v0.2.30` to `v0.2.36` in `_build.yml`.

The release matrix already targets `cuda-version: "13.2.0"`, but `v0.2.30` doesn't know about
that version (CUDA 13.2.0 links were only added to the action in `v0.2.32`), so every `cu132`
build job fails with `Error: Version not available: 13.2.0`, regardless of torch version. This
surfaced during the `v2.3.2.post4` release build. `v0.2.36` is the latest release and includes
the missing version links.


ajf-codes:
again, this was supposed to be against a separate fork of `mamba`. disregard this PR

PR #1033, opened 2026-09-03
[1 of 5] Sequences longer than ~61K tokens silently corrupt Mamba2's chunk state

chipsncheese-ie:
Fixes part of #1015.

Triton specializes pointer-arithmetic operands as 32-bit at kernel-specialization time, so a chained offset multiplication can silently wrap past `2**31 - 1` even though it should be computed in 64-bit — causing wrong output or an illegal memory access on long sequences / large batches.

Casts the relevant program-id index to `tl.int64` in every kernel in `ssd_chunk_state.py`. Tests added for `_chunk_cumsum_fwd` and `chunk_state_varlen`, verified to fail without the fix.


chipsncheese-ie:
Closing and reopening as a fresh PR to avoid the cross-references to the closed stack PRs.

PR #1034, opened 2026-09-03
[2 of 5] Large batches silently corrupt Mamba2's B/C chunk projections

chipsncheese-ie:
Fixes part of #1015.

Same overflow class described in #1015: a chained pointer-arithmetic offset that Triton specializes as 32-bit silently wraps past `2**31 - 1` on large batches, instead of being computed in 64-bit.

Casts the relevant program-id index to `tl.int64` in both kernels in `ssd_bmm.py`. Test added for `_bmm_chunk_fwd`'s batch-axis overflow, verified to fail without the fix.


chipsncheese-ie:

PR #1036, opened 2026-09-03
[4 of 5] Long sequences silently corrupt Mamba2's backward-pass gradients

chipsncheese-ie:
Fixes part of #1015.

Same overflow class described in #1015. Also drops an unconditional `dout.contiguous()` coercion in `_mamba_chunk_scan_combined_bwd` that was silently masking the exact non-contiguous-dout scenario this bug shows up in.

Casts the relevant program-id index to `tl.int64` in every kernel in `ssd_chunk_scan.py` and in the shared backward kernel in `ssd_combined.py`. Tests added for `mamba_chunk_scan_combined`'s forward/backward paths, verified to fail without the fix.


chipsncheese-ie:

PR #1037, opened 2026-09-03
[5 of 5] Large inputs silently corrupt Mamba2's causal_conv1d addressing

chipsncheese-ie:
Fixes part of #1015.

`causal_conv1d` addresses its input with `uint32_t` strides, so `ensure_stride` needs to check against `_UINT32_MAX`, not just the multiple-of-8 stride requirement it already had. Without this, a wide non-contiguous view can reach `causal_conv1d` with a max offset past what it can safely address.

Warns and falls back to `.contiguous()` when that limit would be exceeded, mirroring the existing multiple-of-8 check. Tests added for `causal_conv1d_bwd_dx_given` and `mamba_split_conv1d_scan_combined`, verified to fail without the fix.


chipsncheese-ie:

PR #1038, opened 2026-09-03
[1 of 5] Sequences longer than ~61K tokens silently corrupt Mamba2's chunk state

chipsncheese-ie:
Fixes part of #1015.

Triton specializes pointer-arithmetic operands as 32-bit at kernel-specialization time, so a chained offset multiplication can silently wrap past `2**31 - 1` even though it should be computed in 64-bit — causing wrong output or an illegal memory access on long sequences / large batches.

Casts the relevant program-id index to `tl.int64` in every kernel in `ssd_chunk_state.py`. Tests added for `_chunk_cumsum_fwd` and `chunk_state_varlen`, verified to fail without the fix.

First of a stack of small PRs splitting up #1015's fix by file for easier review.

### DIFF
--- a/mamba_ssm/ops/triton/ssd_chunk_state.py
+++ b/mamba_ssm/ops/triton/ssd_chunk_state.py
@@ -54,8 +54,9 @@ def _chunk_cumsum_fwd_kernel(
     HAS_DT_BIAS: tl.constexpr,
     BLOCK_SIZE_H: tl.constexpr, BLOCK_SIZE_CHUNK: tl.constexpr,
 ):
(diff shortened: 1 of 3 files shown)


KyleMylonakisProtopia:
This is important to see landed. Its fixing a out of bounds run that can result in crashes and correctness of systems using state of the art open weight models like Nemotron Ultra, which have been trained and advertised to use contexts that far exceed where this bug has been shown to occur.

KyleMylonakisProtopia (approved):

qthequartermasterman:
@tridao Is there any update on this?

Issue #1022, opened 2026-08-23
Mamba-3 incremental decode broken: `step()` receives `(B, 1, D)` but `_preprocess` expects 2-D — EinopsError in `.generate()` for both official 1.5B checkpoints


robertforbes68:
## Environment

- GPU: NVIDIA GB10 (DGX Spark, sm_121), aarch64
- CUDA 13.0, torch 2.11.0+cu130, Python 3.12
- mamba-ssm: built from git main on 2026-08-22 (`--no-build-isolation`)
- Checkpoints: `state-spaces/mamba3-siso-1.5b` and `state-spaces/mamba3-mimo-1.5b` (unmodified)
- Tokenizer: Llama-3 (`NousResearch/Meta-Llama-3-8B`)

## Repro

```python
import torch
from mamba_ssm.models.mixer_seq_simple import MambaLMHeadModel
from transformers import AutoTokenizer

tok = AutoTokenizer.from_pretrained("NousResearch/Meta-Llama-3-8B")
model = MambaLMHeadModel.from_pretrained(
    "state-spaces/mamba3-siso-1.5b", device="cuda", dtype=torch.bfloat16)
ids = tok("Warm up.", return_tensors="pt").input_ids.to("cuda")
model.generate(input_ids=ids, max_length=ids.shape[1] + 8, cg=False,
               temperature=1.0, top_k=1)   # same failure with cg=True
```

## Result

```
  File ".../mamba_ssm/modules/mamba3.py", line 172, in forward
    out, _, _, _, _ = self.step(u, angle_dt_state, ssm_state, k_state, v_state)
  File ".../mamba_ssm/modules/mamba3.py", line 354, in step
    DT, B, C, x, z, trap, A, angles = self._preprocess(
  File ".../mamba_ssm/modules/mamba3.py", line 288, in _preprocess
    B = rearrange(B, "b (r g s) -> b r g s", g=self.num_bc_heads, r=rank)
einops.EinopsError: Error while processing rearrange-reduction pattern "b (r g s) -> b r g s".
 Input tensor shape: torch.Size([1, 1, 128]). Additional info: {'g': 1, 'r': 1}.
 Wrong shape: expected 2 dims. Received 3-dim tensor.
```

Identical failure on the MIMO checkpoint (shape `[1, 1, 512]`, `r=4`) and with `cg=True`
(fails inside `update_graph_cache` → `capture_graph`).

## Analysis

[...]

Issue #1014, opened 2026-08-08
Mamba3.forward() passes a 3-D tensor to step(), which requires 2-D — decode path raises EinopsError


JuiceB0xC0de:
`Mamba3.forward()` routes to `self.step(...)` when
`inference_params.seqlen_offset > 0`, but passes `u` through unchanged. `u` is
`(batch, seqlen, hidden_dim)` at that point, while `step()` documents and
requires `(batch, d_model)`. The sequence axis is never squeezed, so the
cached-decode path raises before any kernel runs.

## Where

`mamba_ssm/modules/mamba3.py`, in `forward()`:

```python
def forward(self, u, seq_idx=None, cu_seqlens=None, inference_params=None):
    """
    u: (batch, seqlen, hidden_dim)
    Returns: same shape as u
    """
    batch, seqlen, dim = u.shape          # u is 3-D here
    ...
    if inference_params.seqlen_offset > 0:
        out, _, _, _, _ = self.step(u, angle_dt_state, ssm_state, k_state, v_state)
        return out
```

and in `step()`, same file:

```python
    Args:
        u: (batch, d_model)               # 2-D
```

The contract mismatch is between two methods of the same class.

## Traceback

Observed with `mamba-ssm 2.3.2.post1`, torch 2.11.0+cu128, CUDA 12.8, on
`state-spaces/mamba3-siso-187m`
(`6792c27c00f3bb41506db1066dcd1c51bb0f4b02`), driving one token at a time with
`inference_params.seqlen_offset > 0`:

```
File "mamba_ssm/modules/mamba3.py", line 172, in forward
    out, _, _, _, _ = self.step(u, angle_dt_state, ssm_state, k_state, v_state)
File "mamba_ssm/modules/mamba3.py", line 354, in step
    DT, B, C, x, z, trap, A, angles = self._preprocess(
File "mamba_ssm/modules/mamba3.py", line 288, in _preprocess
    B = rearrange(B, "b (r g s) -> b r g s", g=self.num_bc_heads, r=rank)
einops.EinopsError:  Error while processing rearrange-reduction pattern
  "b (r g s) -> b r g s".
 Input tensor shape: torch.Size([1, 1, 128]).
 Additional info: {'g': 1, 'r': 1}.
 Wrong shape: expected 2 dims. Received 3-dim tensor.
```

`B` is `(batch, seqlen, d_state)` because `self.in_proj(u)` was given a 3-D
[...]

PR #1039, opened 2026-09-04
fix: accept (B, 1, D) in Mamba3.step like Mamba2

Chessing234:
## Summary
- `Mamba3.forward()` routes `inference_params.seqlen_offset > 0` into `step()` with `u` still shaped `(B, L, D)`.
- `step()` only handled `(B, D)`, so decode / `.generate()` raised `EinopsError` in `_preprocess` (issues #1014 / #1022).
- Mirror `Mamba2.step`: accept `(B, 1, D)`, squeeze for the kernel path, unsqueeze the returned token so callers keep their input rank.

## Test plan
- [x] Added `tests/modules/test_mamba3_step_shape.py` (stubs CuteDSL kernel; checks `(B, D)` and `(B, 1, D)` round-trip shapes)
- [ ] On CUDA with CuteDSL deps: `model.generate(...)` for an official Mamba-3 checkpoint past the first token

### DIFF
--- a/mamba_ssm/modules/mamba3.py
+++ b/mamba_ssm/modules/mamba3.py
@@ -320,20 +320,26 @@ class Mamba3(nn.Module):
         will be made available in the future.
 
         Args:
-            u: (batch, d_model)
+            u: (batch, d_model) or (batch, 1, d_model) — same decode contract as Mamba2.step
             angle_state: (batch, nheads, num_rope_angles)
             ssm_state: (batch, nheads, headdim, d_state)
             k_state: (batch, R, nheads, d_state), where R = mimo_rank (R=1 if not MIMO)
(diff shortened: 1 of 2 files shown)


imao81:
Heads-up: I've run real-hardware verification on this branch — my 6-item contract test suite passes 6/6 (SISO/MIMO × 3D decode, bitwise 2D parity, decode-vs-forward within 2e-2), and generate() with the official mamba3-siso-1.5b checkpoint works end-to-end. Full evidence and the test commit are in Chessing234/mamba#1 (a PR onto this branch).
One issue: the CPU test included here will fail as-is , cause it doesn't stub the Triton versions of B_norm/C_norm. Happy to push a fix for that into Chessing234/mamba#1 if you want.

PR #651, opened 2024-12-18
Add selective_scan compilable/exportable custom_ops

bhack:

PR #1042, opened 2026-09-10
Migrate selective scan to the PyTorch stable ABI

### LINKED ISSUE #813: Make mamba libtorch ABI stable
Now that pytorch offers a decent subset of stable APIs, it should be relatively straightforward to have mamba be ABI stable with libtorch by completing the following steps:

1. Switch from using PYBIND to TORCH_LIBRARY (this may also allow mamba to be CPython agnostic) in https://github.com/state-spaces/mamba/blob/main/csrc/selective_scan/selective_scan.cpp. For an example on how to use TORCH_LIBRARY: https://docs.pytorch.org/tutorials/advanced/cpp_custom_ops.html#defining-an-operator
2. Go through [selective_scan.cpp (https://github.com/state-spaces/mamba/blob/main/csrc/selective_scan/selective_scan.cpp) and replace its torch APIs with versions in `torch/headeronly` and `torch/csrc/stable`, similar to what is done in https://github.com/Dao-AILab/flash-attention/pull/1791.

I'm opening this issue to see if this would be of interest to the community + to start a discussion!

cc @mikaylagawarecki

stmcgovern:
Closes https://github.com/state-spaces/mamba/issues/813

This PR is an effort in the direction of providing a stable pytorch c++ extension ecosystem. This could help clear up some of the pytorch version build issues.

Rewrite selective_scan.cpp and the CUDA kernel headers to use only torch/csrc/stable/, torch/headeronly/ and cuda_runtime.h. No c10/, ATen/ or torch/csrc/api/ includes remain, so the extension no longer depends on libtorch's C++ symbols and stays loadable across torch versions.

C++:
- Replace PYBIND11_MODULE with STABLE_TORCH_LIBRARY + TORCH_BOX
- Replace at::Tensor with torch::stable::Tensor, passed by const-ref
- Replace c10::cuda::CUDAGuard/CUDAStream with the stable accelerator API
- Replace AT_ERROR/TORCH_CHECK with STD_TORCH_CHECK
- Replace C10_CUDA_CHECK with STD_CUDA_CHECK
- Replace ATen/cuda/Atomic.cuh with an inline gpuAtomicAdd
- Replace at::Half/at::BFloat16 with their torch::headeronly equivalents
- Use a std::conj replacement for complex conj in device code

Python:
- Dispatch through torch.ops.selective_scan.fwd/bwd, cached at import, since the kernel is now registered as a dispatcher op rather than exposed as pybind11 module functions

Build:
- Add -DTORCH_STABLE_ONLY to the host compile and -DUSE_CUDA to nvcc
- Link with --as-needed

The -std= flags are deliberately left at their existing values. This translation unit no longer includes ATen, so ATen's "C++20 or later compatible compiler is required" guard does not apply to it, and the standard question is being settled separately in
https://github.com/state-spaces/mamba/pull/1000.

Verified against torch 2.15.0a0+git560a746 / CUDA 12.9 on an H100: builds clean, and the suite reports 672 passed with a failure and error set identical to main's baseline on the same machine. The built extension has no undefined at::, c10:: or torch:: C++ symbols; all 73 torch symbols it imports are C shim entry points.

torch.compile behaviour is unchanged from main: the default path (fullgraph=False) works on both, and fullgraph=True fails on both -- on main because the pybind11 call cannot be traced, here because the op has no registered fake implementation. Making fullgraph=True work is a new capability the dispatcher registration makes possible, and is left to a follow-up rather than folded in here.

AI assistance was used to author this code.


stmcgovern:
cc @cleonard530 @janeyx99

cleonard530:
On a quick look, I see that there is no `CMakeLists.txt` file, but I think there needs to be some way to ensure a `TORCH_TARGET_VERSION` flag is used in the build and the `torch` version will need to be guarded in the `pyproject.toml`. I quickly saw that you are using `(cudaStream_t)torch::stable::accelerator::getCurrentStream(device_index).nativeHandle();` which was added in torch 2.13, so you will need `TORCH_TARGET_VERSION=2.13` and `torch>=2.13` at a minimum. 

If you `TORCH_TARGET_VERSION` is not chosen explicitly, then `TORCH_FEATURE_VERSION` will default to the `TORCH_ABI_VERSION` (the current headers torch version)

csrc/selective_scan/selective_scan.cpp
```diff
@@ -2,52 +2,52 @@
...
-        __VA_ARGS__();                                                              \
-    } else if (ITYPE == at::ScalarType::Float)  {                                   \
-        using input_t = float;                                                      \
-        __VA_ARGS__();                                                              \
-    } else {                                                                        \
-        AT_ERROR(#NAME, " not implemented for input type '", toString(ITYPE), "'"); \
-    }
+static inline cudaStream_t get_cuda_stream(int32_t device_index) {
+    return (cudaStream_t)torch::stable::accelerator::getCurrentStream(
+        device_index).nativeHandle();
```
cleonard530: This needs `TORCH_FEATURE_VERSION >= TORCH_VERSION_2_13_0`

stmcgovern: thanks!

csrc/selective_scan/selective_scan.cpp
```diff
@@ -268,109 +268,105 @@ selective_scan_fwd(const at::Tensor &u, const at::Tensor &delta,
...
+        STD_TORCH_CHECK(z.stride(-1) == 1 || z.size(-1) == 1);
         CHECK_SHAPE(z, batch_size, dim, seqlen);
-        out_z = torch::empty_like(z);
+        out_z = torch::stable::empty_like(z);
     }
 
     const int n_chunks = (seqlen + 2048 - 1) / 2048;
-    // const int n_chunks = (seqlen + 1024 - 1) / 1024;
-    // at::Tensor out = torch::empty_like(u);
-    // Right now u has BHL layout and delta has HBL layout, and we want out to have HBL layout
```
cleonard530, line 310: I would probably leave the comments in there as long as they are still relevant, and just updated them to use torch::stable syntax.

csrc/selective_scan/selective_scan.cpp
```diff
@@ -268,109 +268,105 @@ selective_scan_fwd(const at::Tensor &u, const at::Tensor &delta,
...
-    at::Tensor out = torch::empty_like(delta);
-    at::Tensor x;
-    x = torch::empty({batch_size, dim, n_chunks, dstate * 2}, u.options().dtype(weight_type));
+    Tensor out = torch::stable::empty_like(delta);
+    Tensor x = torch::stable::new_empty(u, {batch_size, dim, n_chunks, dstate * 2}, weight_type);
 
     SSMParamsBase params;
     set_ssm_params_fwd(params, batch_size, dim, seqlen, dstate, n_groups, n_chunks, is_variable_B, is_variable_C,
                        u, delta, A, B, C, out, z, out_z,
-                       D_.has_value() ? D_.value().data_ptr() : nullptr,
```
cleonard530, line 318: Was this a bug before? You are using `D` which is only set if `D_.has_value()` is true, so practically these should be the same but it might of been a bug before where `D_` was just assumed to always have a value?

csrc/selective_scan/selective_scan.cpp
cleonard530: No need to update this, but if this it the only thing requiring torch version >= 2.13, you can check out how vLLM used implemented this with `get_current_cuda_stream` here https://github.com/vllm-project/vllm/blob/main/csrc/libtorch_stable/torch_utils.h

stmcgovern: yes the min pytorch version needed is really only 2.10. this one call can be worked around.

csrc/selective_scan/selective_scan_common.h
cleonard530, line 25: I got a build error for std::max. I think the old header pulled in `<algorithm>` through transitive imports but the new header does not. According to codex, you may also need `<initializer_list>`

csrc/selective_scan/selective_scan_bwd_bf16_complex.cu
```diff
@@ -6,4 +6,4 @@
 
 #include "selective_scan_bwd_kernel.cuh"
 
-template void selective_scan_bwd_cuda<at::BFloat16, complex_t>(SSMParamsBwd &params, cudaStream_t stream);
\ No newline at end of file
+template void selective_scan_bwd_cuda<c10::BFloat16, complex_t>(SSMParamsBwd &params, cudaStream_t stream);
```
cleonard530: why are all of these changed to `c10` and not `torch::headeronly`?

cleonard530:
The library looks to be stable now but there are a lot of references to `c10`. These should be replaced by `torch::headeronly` to guarantee that they stay independent of libtorch. Let me know if there are any references to `c10` that do not have a `torch::headeronly` alias.

csrc/selective_scan/selective_scan_common.h
cleonard530: After I added these headers the build succeeded.

csrc/selective_scan/selective_scan.cpp
stmcgovern: restored

csrc/selective_scan/selective_scan.cpp
stmcgovern: reverted.

csrc/selective_scan/selective_scan_common.h
stmcgovern: headers added

csrc/selective_scan/selective_scan_bwd_bf16_complex.cu
stmcgovern: fixed

csrc/selective_scan/selective_scan.cpp
```diff
@@ -388,82 +412,82 @@ selective_scan_bwd(const at::Tensor &u, const at::Tensor &delta,
...
-    at::Tensor dA = torch::zeros_like(A);
-    at::Tensor dB = !is_variable_B ? torch::zeros_like(B) : torch::zeros_like(B, B.options().dtype(torch::kFloat32));
-    at::Tensor dC = !is_variable_C ? torch::zeros_like(C) : torch::zeros_like(C, C.options().dtype(torch::kFloat32));
-    at::Tensor dD;
-    if (D_.has_value()) { dD = torch::zeros_like(D_.value()); }
-    at::Tensor ddelta_bias;
-    if (delta_bias_.has_value()) { ddelta_bias = torch::zeros_like(delta_bias_.value()); }
+    Tensor du = torch::stable::empty_like(u);
+    Tensor ddelta = torch::stable::empty_like(delta);
+    Tensor dA = torch::stable::new_zeros(A, A.sizes());
```
cleonard530, line 484: Can you check to see if this produces the same thing as zeros_like, particularly with things like layout and device? I am not sure if new_zeros inherits this from A or not, and there may be issues with non-contiguous tensors.

csrc/selective_scan/selective_scan.cpp
```diff
@@ -2,52 +2,77 @@
...
-    } else if (ITYPE == at::ScalarType::BFloat16) {                                 \
-        using input_t = at::BFloat16;                                               \
-        __VA_ARGS__();                                                              \
-    } else if (ITYPE == at::ScalarType::Float)  {                                   \
-        using input_t = float;                                                      \
-        __VA_ARGS__();                                                              \
-    } else {                                                                        \
-        AT_ERROR(#NAME, " not implemented for input type '", toString(ITYPE), "'"); \
-    }
+#define CHECK_SHAPE(x, ...) STD_TORCH_CHECK(x.sizes() == torch::headeronly::IntHeaderOnlyArrayRef({__VA_ARGS__}), #x " must have shape (" #__VA_ARGS__ ")")
```
cleonard530, line 37: I think there might be an issue here as well. I've had to change this to .equals() instead of == before. I think == was added for stable tensors in a later version.

cleonard530:
The README still says `PyTorch 1.12+` so you should update that as well. After that and the comments I left above it looks good to me! I would maybe suggest building on one torch version and then running on another to make sure it works.

csrc/selective_scan/selective_scan.cpp
```diff
@@ -478,20 +517,37 @@ selective_scan_bwd(const at::Tensor &u, const at::Tensor &delta,
...
 
-PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
-    m.def("fwd", &selective_scan_fwd, "Selective scan forward");
-    m.def("bwd", &selective_scan_bwd, "Selective scan backward");
+STABLE_TORCH_LIBRARY(selective_scan, m) {
+    m.def("fwd(Tensor u, Tensor delta, Tensor A, Tensor B, Tensor C, "
+          "Tensor? D, Tensor? z, Tensor? delta_bias, bool delta_softplus) -> Tensor[]");
+    m.def("bwd(Tensor u, Tensor delta, Tensor A, Tensor B, Tensor C, "
+          "Tensor? D, Tensor? z, Tensor? delta_bias, Tensor dout, "
+          "Tensor? x, Tensor? out, Tensor? dz, "
```
cleonard530, line 538: Is `dz` read-only or does the bwd kernel write into it at line 210 in [selective_scan/selective_scan_bwd_kernel.cuh‎](https://github.com/state-spaces/mamba/pull/1042/changes#diff-e3dfdeb30e92bad339ac996110b4fa0de233285c57d6e6e84700c42ee86f8f5a)?

csrc/selective_scan/selective_scan.cpp
cleonard530, line 141: replace data pointers with either `mutable_data_ptr()` or `const_data_ptr()`

cleonard530:
@stmcgovern, the build worked on my end but my agent caught a few more issues (see comments above). The one about `dz` is definitely an issue if it is true, so please check this.

PersusUS:
Answering @cleonard530's question about `dz`: the kernel writes into it, and the caller depends on that write.

`selective_scan_bwd_kernel.cuh:210` does `store_output<Ktraits>(dz, dz_vals, smem_store, ...)`, where `dz` is `params.dz_ptr`, set from the caller-supplied tensor whenever `dz_.has_value()`. It is an output buffer, not a read-only input.

And the caller relies on that. In `MambaInnerFn.backward`:

```python
dxz = torch.empty_like(xz)          # uninitialised
dx, dz = dxz.chunk(2, dim=1)        # dz is a view into dxz
... = _selective_scan_bwd(..., dz, ...)
...
return (dxz, ...)                   # the z half is only ever filled by that write
```

So the schema in this PR is inaccurate:

```
m.def("bwd(... Tensor? x, Tensor? out, Tensor? dz, ...) -> Tensor[]");
```

`Tensor? dz` declares a read-only input. It needs to be annotated as mutated, and the returned `dz` (pushed into `result` when `has_z`) aliases that same input, so the alias needs expressing too — roughly `Tensor(a!)? dz` plus the matching annotation on the output.

This was harmless under the old `PYBIND11_MODULE` binding, which carried no schema at all, and it stays harmless in eager mode. It stops being harmless now that this is a real `STABLE_TORCH_LIBRARY` op, because functionalization takes the schema at its word: an undeclared input mutation can land in a clone rather than the caller's buffer, and then `dxz[:, dim:]` keeps whatever `torch.empty_like` happened to return. Silent, gradient-valued and nondeterministic.

`torch.library.opcheck` should catch it — `test_aot_dispatch_dynamic` is the one that exercises functionalization.

`fwd` looks clean by the same reading: it allocates all of its outputs and takes no caller-provided buffer.

A second, smaller point, in case it is in scope here. Only `STABLE_TORCH_LIBRARY_IMPL(selective_scan, CUDA, m)` is registered, with no `Meta`/fake implementation. With the op registered but no fake kernel, `torch.compile(fullgraph=True)` still cannot trace through `selective_scan`. That is the same gap #651 was closing from the `torch.library.custom_op` side, so the two changes look complementary rather than competing.

For context on why I care: I ran into the fullgraph limitation last spring while entering a Mamba/Transformer hybrid into OpenAI's Parameter Golf, and worked around it by rewriting the recurrence as an associative scan in pure PyTorch. I later wrote the same recurrence as CUDA kernels ([selective-scan-cuda](https://github.com/PersusUS/selective-scan-cuda)), so if a second set of numbers is useful I am happy to test this PR on an Ada laptop GPU (sm_89), including the build-on-one-torch-version, run-on-another check you asked for.

stmcgovern:
> This was harmless under the old `PYBIND11_MODULE` binding, which carried no schema at all, and it stays harmless in eager mode. It stops being harmless now that this is a real `STABLE_TORCH_LIBRARY` op, because functionalization takes the schema at its word:

Thanks @PersusUS ! For `dz`, there's no fake implementation added in this PR, so under compile this will cause a graph break and the eager behavior is unchanged. I completely agree that the schema needs to be changed for correct functionalization. My thinking on the current PR is to keep the focus on moving to the stable abi, trying to keep strict parity with the behavior on main. So my inclination is not to modify the function signature in this PR, leaving that to a follow up. 

IMO modifying the schema and adding a meta impl would be nice as a feature addition to allow for compile with fullgraph=True . Up to the maintainers, of course to decide on that. 

More testing of this PR on different hardware and pytorch versions would be very welcome!

csrc/selective_scan/selective_scan.cpp
stmcgovern: fixed

csrc/selective_scan/selective_scan.cpp
stmcgovern: changed

csrc/selective_scan/selective_scan.cpp
stmcgovern: done

csrc/selective_scan/selective_scan.cpp
stmcgovern: this was addressed in comments below.

cleonard530: I think you still need to replace this with something like `Tensor(dz!)? dz` right?

stmcgovern: yes, your'e right. it needs to be marked in-place and optional.

PersusUS:
Correction first: I said an undeclared mutation could end up in a clone and leave the caller's buffer uninitialised. I can't reproduce that. With the old schema the write lands in the caller's buffer either way. I was reasoning from the docs rather than from something I'd run.

The annotation still doesn't look sufficient on its own though, because `bwd` also returns `dz`: `result.push_back(dz)` when `has_z` (selective_scan.cpp:531 as of d9233be). Same tensor that came in, and the schema doesn't declare that, so test_schema fails:

```
Argument dz is not defined to alias output but was aliasing
```

I'm passing `test_utils=["test_schema"]` because plain opcheck also trips test_faketensor here, which I'm assuming is the missing fake impl you mentioned.

It only fires when `z` is used and the caller hands in its own `dz`, which is `MambaInnerFn.backward`: `dxz = torch.empty_like(xz)`, `dx, dz = dxz.chunk(2, dim=1)`, then `dz` goes in at selective_scan_interface.py:349.

Two things make it pass: return `dz.clone()`, or drop `dz` from the returned list. Dropping isn't free though. `MambaInnerFn.backward` passes its own `dz`, a view into `dxz`, so it wouldn't miss it. But `SelectiveScanFn.backward` passes None at selective_scan_interface.py:95 and reads the gradient back out of the result at line 98, so it would have to allocate its own buffer and pass it in, which the comment right above that call says the kernel supports. Either way it's a Python-side change, so follow-up rather than this PR.

Caveat on all of it: I couldn't build the extension, so this isn't the real op. It's a stand-in with the same schema string and the same list layout, torch 2.13.0 on CPU:

```python
lib = torch.library.Library("probe", "DEF")
lib.define("bwd(Tensor u, Tensor delta, Tensor A, Tensor B, Tensor C, "
           "Tensor? D, Tensor? z, Tensor? delta_bias, Tensor dout, "
           "Tensor? x, Tensor? out, Tensor(dz!)? dz, "
           "bool delta_softplus, bool recompute_out_z) -> Tensor[]")

def impl(u, delta, A, B, C, D, z, delta_bias, dout, x, out, dz,
         delta_softplus, recompute_out_z):
    has_z = z is not None
    res = [torch.empty_like(u) for _ in range(7)]
    if has_z:
        if dz is None:
            dz = torch.empty_like(z)
        dz.fill_(8.)
        res.append(dz)                      # result.push_back(dz)
    return res

lib.impl("bwd", impl, "CPU")

n = torch.zeros(4)
torch.library.opcheck(
    torch.ops.probe.bwd,
    (n, n, n, n, n, None, n, None, n, None, None, n.clone(), False, False),
    test_utils=["test_schema"],
)
```

On the extra testing: I offered sm_89 numbers in my first comment and I can't run that on my own machine after all. I can do a T4 on Linux instead if a second set of numbers is useful.

PR #1043, opened 2026-09-11
Deterministic mode: pin tile sizes per kernel from TRITON_AUTOTUNE_BLOCK* variables

NolenLiang:
`_filter_configs_by_block_sizes` reads only `TRITON_AUTOTUNE_BLOCK_SIZE_{M,N,K,DSTATE}` and applies every set variable to every kernel. Two consequences: the chunk-cumsum head tile (`BLOCK_SIZE_H`) cannot be pinned from the environment, so deterministic mode (with cached autotuning off) always takes the cheapest config, `BLOCK_SIZE_H=1`; and a variable naming a key a kernel does not have filters out all of that kernel's configs and silently falls back to the cheapest one. This change reads every `TRITON_AUTOTUNE_BLOCK*` variable and applies, per kernel, only the constraints whose key appears in that kernel's candidate configs (ANDed); keys the kernel does not use are ignored; only when the applicable constraints match nothing does it fall back to the cheapest config. `TRITON_AUTOTUNE_BLOCK_SIZE_H=2` therefore selects the H=2 cumsum config while the other kernels keep their cheapest config, and combinations such as `BLOCK_SIZE_M` plus `BLOCK_SIZE_H` behave as expected. Tests cover H on the cumsum kernel, H on a kernel without that key, M plus H, a constraint that matches nothing, non-integer values (ignored for unrelated keys, an error for applicable ones), deterministic mode off, and cached autotuning on. Motivation: `tl.cumsum` results depend on the head tile, and matching that tile between training (mamba_ssm) and inference (vLLM batch-invariant mode, which pins H=2) makes the SSD forward bitwise identical across the two on the shapes and chunk sizes I measured.


NolenLiang:
@tridao @paul-gibbons could you take a look? This only touches the deterministic-mode config selection: it lets `TRITON_AUTOTUNE_BLOCK_SIZE_H` (and any other `TRITON_AUTOTUNE_BLOCK*` variable) pin a kernel's tile, applied per kernel to the keys its configs actually have. The motivation is matching the chunk-cumsum head tile between training (this package) and vLLM's batch-invariant mode, which pins H=2; details in the description.

